PR #22, opened 2018-12-21
menuconfig: fix inputbox with integer configs

kilroyd:
An inputbox always wants to deal with strings, so make sure we
convert any integer config values for display.

When setting the value of the config make sure we convert back to an
int, and that we won't try to test the integer against a string.

Also split MenuItem.set_bool() into MenuItem.set() and
MenuItem.clear(), so the caller is isolated from how we encode the
boolean internally.

### DIFF
--- a/config_system/general.py
+++ b/config_system/general.py
@@ -461,7 +461,7 @@ def set_config(key, value, is_user_set = True):
         elif config[key]['datatype'] == 'int':
             # Must convert to an integer
             try:
-                int(value)
+                value = int(value)
             except ValueError:
                 logger.warn("Ignoring integer configuration option %s with non-integer value '%s'" % (key, value))
                 return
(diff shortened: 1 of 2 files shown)


config_system/general.py
```diff
@@ -769,26 +770,43 @@ def needs_inputbox(self):
...
-                self.set_bool('y')
+                self.set()
     def get_value(self):
-        return get_config(self.value)['value']
+        """Always return a string representation"""
+        value = get_config(self.value)['value']
+        if get_config(self.value)['datatype'] == 'int':
+            value = str(value)
+        return value
     def set_value(self, new_value):
```
chrisdiamand: What's stopping someone calling this on an int option?

There's also no set_int function - should there be? Otherwise it seems like you have to use set_value for both? (which would be incorrect, because it would store the menuconfig input text as a string, not an int)

Apologies if I've misunderstood :)

kilroyd: That's right, get_value and set_value are intended to work with any type (int, bool, string) of option.
get_value is generally called to get the text for an input box. And set_value would be called to set the value after the inputbox is closed. Since an inputbox is not used to set the value on booleans, they should only see an int or string in this use case. The other usage is for some logging a little below...

PR #23, opened 2018-12-21
Fix append() usage with toolchain flags

chrisdiamand:
Toolchain flags are getting corrupted because of incorrect usage of
append(). The issue is as follows:

 (1) Toolchain initialisation generates cflags and cxxflags.
 (2) cxxflags is then merged with cflags:
         cxxflags = append(cflags, cxxflags)
 (3) In CompileObjs(), cflags is merged with conlyflags:
         ...append(cctargetflags, Properties.conlyflags)

The issue is that, because of (2), cxxflags ends up using the same
buffer as cflags. This means that the last two elements of that
buffer can be overwritten with conlyflags by (3), with the result
that conlyflags end up in cxxflags!

Fix this in two ways:
 * Swap the order of cflags and cxxflags in (2), so that the cflags
   are copied into cxxflags' buffer.
 * Instead of appending arrays then passing them to strings.Join(),
   add a new helper, utils.Join(), which can join multiple string
   arrays into a new destination string without appending any of the
   source arrays together. This prevents the overwriting in (3).

Change-Id: I652578bdda4cb7ec9cc33cccb97affa4dd44cc19

### DIFF
--- a/core/linux.go
+++ b/core/linux.go
@@ -272,10 +272,10 @@ func (l *library) CompileObjs(ctx blueprint.ModuleContext) []string {
 	cc, cctargetflags := tc.getCCompiler()
 	cxx, cxxtargetflags := tc.getCXXCompiler()
 
-	ctx.Variable(pctx, "asflags", strings.Join(append(astargetflags, l.Properties.Asflags...), " "))
-	ctx.Variable(pctx, "cflags", strings.Join(cflagsList, " "))
-	ctx.Variable(pctx, "conlyflags", strings.Join(append(cctargetflags, l.Properties.Conlyflags...), " "))
-	ctx.Variable(pctx, "cxxflags", strings.Join(append(cxxtargetflags, l.Properties.Cxxflags...), " "))
+	ctx.Variable(pctx, "asflags", utils.Join(astargetflags, l.Properties.Asflags))
+	ctx.Variable(pctx, "cflags", utils.Join(cflagsList))
+	ctx.Variable(pctx, "conlyflags", utils.Join(cctargetflags, l.Properties.Conlyflags))
+	ctx.Variable(pctx, "cxxflags", utils.Join(cxxtargetflags, l.Properties.Cxxflags))
 
 	var objectFiles []string
 	for _, source := range srcs {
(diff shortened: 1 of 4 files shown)


core/linux.go
```diff
@@ -272,10 +272,10 @@ func (l *library) CompileObjs(ctx blueprint.ModuleContext) []string {
...
-	ctx.Variable(pctx, "asflags", strings.Join(append(astargetflags, l.Properties.Asflags...), " "))
-	ctx.Variable(pctx, "cflags", strings.Join(cflagsList, " "))
-	ctx.Variable(pctx, "conlyflags", strings.Join(append(cctargetflags, l.Properties.Conlyflags...), " "))
-	ctx.Variable(pctx, "cxxflags", strings.Join(append(cxxtargetflags, l.Properties.Cxxflags...), " "))
+	ctx.Variable(pctx, "asflags", utils.Join(astargetflags, l.Properties.Asflags))
+	ctx.Variable(pctx, "cflags", utils.Join(cflagsList))
+	ctx.Variable(pctx, "conlyflags", utils.Join(cctargetflags, l.Properties.Conlyflags))
+	ctx.Variable(pctx, "cxxflags", utils.Join(cxxtargetflags, l.Properties.Cxxflags))
 
 	var objectFiles []string
```
kilroyd: Please fix the strings.Join(append()) in getCommonLibArgs as well (~line 529)

utils/utils.go
```diff
@@ -190,3 +190,22 @@ func Trim(args []string) []string {
 	}
 	return out
 }
+
+// Join multiple lists of strings. This replaces appending multiple arrays
+// together before calling strings.Join().
+func Join(lists ...[]string) (joined string) {
+	first := true
+	for _, list := range lists {
+		for _, elem := range list {
```
kilroyd: Should we be calling Strings.Join to join the content of each slice?
I guess we want to avoid re-allocating string buffers, but here we're constantly extending the string buffer for every element

chrisdiamand: Good question...

I originally did it this way so that it would be easy to convert to use a Builder: https://golang.org/pkg/strings/#Builder
That would be possible to do if we eventually require Go 1.10.

But you're right, it would be faster for now to use strings.Join() on each element. It might be even better to count the total length at the start though and allocate the correct length for the entire final string - I'll try it and see. That's also probably not much worse than using a Builder.

PR #34, opened 2019-01-18
Add check-signoff

gelldur:
Improve Travis check scripts and add checking for sign-off

Current solution isn't perfect, has some limitations.

Change-Id: I7ebe50e3be18b7fa7f885c4bce2f96066c97f42c
Signed-off-by: Dawid Drozd <dawid.drozd@arm.com>


.travis/checks/check-signoff.py
```diff
@@ -0,0 +1,43 @@
...
+
+if os.environ.get('CI') and os.environ.get('TRAVIS'):
+	commit_range = os.environ['TRAVIS_COMMIT_RANGE']
+
+print("checking commit_range={}".format(commit_range))
+
+try:
+	# For Travis variables check: https://docs.travis-ci.com/user/environment-variables/
+	# TRAVIS_COMMIT_RANGE - when force update it will contain previous old commit sha eg. 4b7d3xxxxxx...cf1b4xxxxxx
+	# cf1b4xxxxxx - forced pushed commit (after ammend)
```
chrisdiamand: s/ammend/amend

.travis/checks/check-signoff.py
```diff
@@ -0,0 +1,43 @@
...
+	# so commands like git log wil fail in such cases, because 4b7d3xxxxxx doesn't exist in clean copy of repository
+	git_shas = subprocess.check_output(['git', 'log', '--pretty=%H', commit_range]).decode("utf-8")
+	git_shas = git_shas.split('\n')
+	for sha in git_shas:
+		sha = sha.strip()
+		if len(sha) == 0:
+			continue
+		# Extract every commit message one by one. Maybe in commit someone placed some code or other odd chars.
+		message = subprocess.check_output(['git', 'log', '--pretty=%B', '-n1', sha]).decode("utf-8")
+		if "Signed-off-by" not in message:
```
chrisdiamand: Do we want to ensure that the Signed-off-by is at the start of a line? At the moment, this will match anywhere in the commit message. Maybe something like:

    for line in message.splitlines():
        if line.startswith("Signed-off-by: "):
            found_signoff = True

?

gelldur: Little bit "raw" but I think it would be better.

PR #35, opened 2019-01-21
Fix Python 3 issues

chrisdiamand:
Fix two issues preventing Python 3 from being usable:
 - Hash functions need to be fed bytes, not strings
 - Use the `print()` function

Update Travis to test with Python 3 as well as 2.7. Unfortunately,
Travis cannot support multiple languages with the `language`
directive, so instead use an environment variable set from a matrix
to choose the version.

Change-Id: Ica35e4a0e22c8c4cb14efa20ede264fe0f19d5cd

### DIFF
--- a/.travis/run_all_tests.sh
+++ b/.travis/run_all_tests.sh
@@ -3,6 +3,12 @@ source .travis/utils.sh
 STATUS_CODE=0 # reset
 
 ####################
+fold_start 'setup_python'
+    source ${BOB_ROOT}/.travis/set_python_version.sh
+    check_result $? "Check setup_python"
+fold_end 'setup_python'
+
+####################
 fold_start 'run_build_tests.sh'
     bash ${BOB_ROOT}/.travis/run_build_tests.sh
     check_result $? "Check run_build_tests: "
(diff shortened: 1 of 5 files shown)


.travis/run_all_tests.sh
```diff
@@ -2,6 +2,41 @@
 source .travis/utils.sh
 STATUS_CODE=0 # reset
 
+set -ex
+
+####################
+fold_start 'setup_python'
```
gelldur: Maybe this should be moved to init part (before script in Travis) because run_all_tests should be running tests only no setup here. Also it would be nic to enclose all those commands in 1 nice bash script (less code here)

gelldur: Also If you move to separate script you can just call `exit 1` etc

chrisdiamand: Yep - I can move to a separate script, but it has to be `source`d because it's modifying PATH.

gelldur: Yeah and this is why I mean to run it directly from .travis.yml

PR #39, opened 2019-01-29
Add test to use `generated_headers` and `generated_sources` on the same target

gelldur:
Change-Id: Ida0bf8afb9b8baad2b54c667d9de068b06f6009f
Signed-off-by: Dawid Drozd <dawid.drozd@arm.com>

### DIFF
--- a/tests/generate_source/build.bp
+++ b/tests/generate_source/build.bp
@@ -1,5 +1,5 @@
 /*
- * Copyright 2018 Arm Limited.
+ * Copyright 2018-2019 Arm Limited.
  * SPDX-License-Identifier: Apache-2.0
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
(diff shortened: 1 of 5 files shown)


tests/generate_source/build.bp
```diff
@@ -189,12 +189,38 @@ bob_binary {
...
         "generate_source_single_dependend_nested",
     ],
+    generated_headers: [
+        "generate_source_single",
+    ],
     srcs: ["main.cpp"],
 }
 
+bob_generate_source {
+    name: "gen_foo_sources",
```
chrisdiamand: Pref: Maybe you could call this something like "gen_sources_and_headers" to better explain what this is testing?

tests/generate_source/lib.c.txt
```diff
@@ -0,0 +1,21 @@
+/*
```
chrisdiamand: Not required, but you could embed this file in `gen.sh` using a heredoc: http://tldp.org/LDP/abs/html/here-docs.html

```shell
cat >$c_src <<EOF
#include <stdio.h>
#include "foo/foo.h"

void foo() {
    printf("foo()\n");
}
EOF
```

chrisdiamand: *and lib.h, too.

tests/generate_source/main.cpp
```diff
@@ -1,3 +1,20 @@
+/*
+ * Copyright 2018-2019 Arm Limited.
```
chrisdiamand: This file isn't modified in this commit, except adding the copyright header - and hasn't been otherwise changed this year, so we can't really put '2019' in the header :)

Could you leave this change out of the commit please?

tests/generate_source/lib.c.txt
gelldur: I don't want such solution because we need to add this to gitignore

tests/generate_source/build.bp
```diff
@@ -189,12 +189,38 @@ bob_binary {
...
+        "foo/src/foo.c",
+        "foo/foo.h",
+    ],
+    export_gen_include_dirs: ["."],
+    tool: "gen.sh",
+    cmd: "${tool} ${gen_dir} ${in}"
+}
+
+bob_binary {
+    name: "foo_generated",
```
chrisdiamand: Pref: Can you rename this too please? Same reasoning as gen_sources_and_headers.

tests/generate_source/lib.c.txt
chrisdiamand: (i.e. generate directly in the build directory directly from the script, so we don't have as  many source files containing ~5 lines of C :) )

gelldur: Still such solution isn't nice for me. Other tests that we have use same approach. Just prefer to keep template files rather than template in source code.

chrisdiamand: OK, sure!

PR #43, opened 2019-02-07
Pep-8 changes for syntax, lex files

mwidera:
Apply some indentation fixes, Replace most/all of single quotes with
double ones
Fix some occurences when double quotes actually broke code
Fix ranges which were unnecessary changed

Change-Id: I17bbc7b2c1b446905d07b185c3750027ef1a8b05

### DIFF
--- a/config_system/lex.py
+++ b/config_system/lex.py
@@ -16,80 +16,86 @@
 import ply.lex as lex
 import re
 
+
 class TokenizeError(Exception):
     pass
 
+
 tokens = (
-        'ANDAND','OROR', 'NOT',
-        'BOOL', 'BOOLEAN',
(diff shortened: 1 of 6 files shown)


config_system/lex.py
```diff
@@ -16,142 +16,156 @@
...
-    t.type = 'NUMBER'
+    r"[0-9]+"
+    t.type = "NUMBER"
     t.value = int(t.value)
     return t
 
+
 def t_PARAM_string(t):
-    r'"[^"]*"'
+    r'"[^\"]*"'
```
chrisdiamand: Did you intend to add the `\` here?

mwidera: Well actually not. I'll remove this escape character

PR #47, opened 2019-02-11
Refactor host exploration

chrisdiamand:
Deduplicate the remaining host_explore code, so that each operation
can be done on an arbitrary target type.

Change-Id: I988dc80c15cbfff79b791fd25633e999412ff7fe

### DIFF
--- a/scripts/host_explore.py
+++ b/scripts/host_explore.py
@@ -49,63 +49,71 @@ def check_output(command, dir=None):
 
     return output
 
+
+def get_cxx_compiler(tgtType):
+    """Parse the config options prefixed by `tgtType` and determine the name of
+    the compiler for that target type, as well as whether it is a cross
+    compiler.
+    """
+    cross_compiler = False
+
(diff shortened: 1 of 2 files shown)


scripts/host_explore.py
```diff
@@ -49,63 +49,71 @@ def check_output(command, dir=None):
...
+    executed.
+    """
+    cxx, cross_compiler = get_cxx_compiler(tgtType)
+
+    if cross_compiler:
+        return []
+
+    stl_path = check_output([cxx, "-print-file-name=libstdc++.so"])
+    stl_dir = os.path.dirname(stl_path)
+    ld_library_path = os.getenv("LD_LIBRARY_PATH", "")
```
mwidera: I'll suggest to inform via key name about the second argument functionality (which is fallback output)

```suggestion
    ld_library_path = os.getenv("LD_LIBRARY_PATH", default="")
```

chrisdiamand: Sure, I can do that.

PR #52, opened 2019-02-15
New test for output verification with library output field usage

mwidera:
Added new test composition to verify output from library and binary for bob

Change-Id: Ifee147bbe140e8caf456adc15b3449974ca5632d

### DIFF
--- a/core/linux.go
+++ b/core/linux.go
@@ -47,6 +47,9 @@ func pathToLibFlag(path string) string {
 	_, base := filepath.Split(path)
 	ext := filepath.Ext(base)
 	base = strings.TrimSuffix(base, ext)
+	if !strings.HasPrefix(base, "lib") {
+		panic(errors.New("Shared library name must start with 'lib' prefix"))
+	}
 	base = strings.TrimPrefix(base, "lib")
 	return "-l" + base
 }
@@ -83,21 +86,26 @@ var copyRule = pctx.StaticRule("copy",
 		Description: "$out",
 	})
 
-type targetableModule interface {
+type singleOutputModule interface {
 	blueprint.Module
+	outputName() string
+}
+
+type targetableModule interface {
+	singleOutputModule
 	getTarget() string
 }
 
 // Where to put generated shared libraries to simplify linking
 // As long as the module is targetable, we can infer the library path
 func getSharedLibLinkPath(t targetableModule) string {
-	return filepath.Join("$BuildDir", t.getTarget(), "shared", t.Name()+".so")
+	return filepath.Join("$BuildDir", t.getTarget(), "shared", t.outputName()+".so")
 }
 
 // Where to put generated binaries in order to make sure generated binaries
 // are available in the same directory as compiled binaries
 func getBinaryPath(t targetableModule) string {
-	return filepath.Join("$BuildDir", t.getTarget(), "executable", t.Name())
+	return filepath.Join("$BuildDir", t.getTarget(), "executable", t.outputName())
 }
 
 // Generate the build actions for a generateSource module and populates the outputs.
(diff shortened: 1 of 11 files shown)


tests/output/verify.py
```diff
@@ -0,0 +1,30 @@
...
+parser = argparse.ArgumentParser(description='Test generator.')
+parser.add_argument('--out')
+parser.add_argument('--expected')
+
+args = parser.parse_args()
+
+if os.path.basename(args.out) != args.expected:
+    print(os.path.basename(args.out))
+    msg = "Output from generation: {} but expected: {}".format(args.out, args.expected)
+    raise Exception(msg)
```
chrisdiamand: Pref: Raising an exception here will mean we get an unnecessary Python backtrace when this test fails.

Instead, can you just do `sys.exit(1)` (and print the message obviously :) )?

core/linux.go
```diff
@@ -47,6 +47,9 @@ func pathToLibFlag(path string) string {
 	_, base := filepath.Split(path)
 	ext := filepath.Ext(base)
 	base = strings.TrimSuffix(base, ext)
+	if !strings.HasPrefix(base, "lib") {
+		panic(errors.New("Shared library need to start with 'lib' prefix"))
```
chrisdiamand: Grammar: s/need/needs

Although maybe "Shared library output must start with 'lib' prefix" would be clearer?

tests/output/libsh/libsh.h
```diff
@@ -0,0 +1 @@
+extern int libshared(void);
```
chrisdiamand: GitHub's showing me a funny symbol saying there's no line at the end of the file here?

(not a *trailing* blank line - a `\n`character after `libshared(void);`)

tests/output/libst/libst.h
```diff
@@ -0,0 +1 @@
+extern int libstatic(void);
```
chrisdiamand: same here

tests/output/verify.py
chrisdiamand: Github's also complaining about the missing `\n` here too :(

core/linux.go
mwidera: Actually it's not only referring to output field but name in general, so I've changed it as you suggested but with minor change

tests/output/verify.py
mwidera: Well I think it's cleaner use dedicated raise exception form than print based traceback?

chrisdiamand: Thanks!

The issue is that we don't need a backtrace. It's the same as e.g. if I do this:
```bash
# cat non_existent_file
cat: non_existent_file: No such file or directory
```
I get a single-line error message - I don't get a backtrace from `cat`, because the problem is with the user/whatever is running `cat`, not `cat` itself.

A backtrace/Python exception is only useful when there is a bug in the Python itself - but this "output from generation..." message is demonstrating that there's an issue with Bob, *not* with `verify.py` :)

core/generated.go
```diff
@@ -153,6 +153,10 @@ func (m *generateCommon) shortName() string {
 	return m.Name()
 }
 
+func (m *generateCommon) outputName() string {
```
kilroyd: I just realised that outputName on generateCommon is actually incorrect, as it shouldn't be supported by generateSource and transformSource (which have multiple outputs). It should really be on generateLibrary in gen_library.go.

core/linux.go
```diff
@@ -86,18 +89,19 @@ var copyRule = pctx.StaticRule("copy",
 type targetableModule interface {
 	blueprint.Module
 	getTarget() string
+	outputName() string
```
kilroyd: I think this targetableModule interface should be redefined to be based on a new interface (instead of blueprint.Module) where it would get outputName.

core/generated.go
mwidera: moved to gen_library.go

core/linux.go
mwidera: Moved to new interface but name of it is still under discussion

mwidera:
Name for new interface changed, change tested locally

PR #53, opened 2019-02-18
Mconfig formatter implementation

mwidera:
Across the whole change here's what's changed:
- Handling for KConfig help msg format (tabs + 2 spaces indent)
- Introduce mconfigfmt formatter script
- Travis script expanded to handle tests for config_menu module

Run_tests.py changes:
- Exit code 1 when any test failed (error checks for travis)

Run_tests_formatter.py changes:
- Compare expected result of formatting for both input files: wrongly
  formatted and properly formatted one
- Exit code 1 when any test failed (error checks for travis)

Change-Id: I2825bbc69bd29c219534855cc8b621f8bd81af89

### DIFF
--- a/.travis/run_tests.sh
+++ b/.travis/run_tests.sh
@@ -1,0 +1,9 @@
+#!/usr/bin/env bash
+set -eE
+trap "echo '<------------- run_tests.sh failed'" ERR
+
+cd ${BOB_ROOT}/config_system/tests
+which python
+
+# Execute run_test
+python run_tests.py
--- a/config_system/tests/formatter/expected.out
+++ b/config_system/tests/formatter/expected.out
@@ -1,0 +1,69 @@
+# Some comment that need to be parsed!
+
+source "foo"
+
+config MODE
+	bool "Mode"
+	default n
+
+
+choice
+	prompt "Choose"
+	default OPTION_A
+
+config OPTION_A
+	bool "A"
+
+config OPTION_B
+	bool "B"
+	depends on MODE
+
+config OPTION_C
+	bool "C"
+	depends on MODE
+
+endchoice
+choice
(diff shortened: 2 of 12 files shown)


.travis/run_tests.sh
```diff
@@ -0,0 +1,9 @@
+#!/usr/bin/env bash
+set -eE
+trap "echo '<------------- run_tests.sh failed'" ERR
+
+cd ${BOB_ROOT}/config_system/tests
+which python
+
+# Execute run_test
+python run_tests.py
```
kilroyd: Fix newline at end of file please

config_system/tests/formatter/expected.out
```diff
@@ -0,0 +1,69 @@
...
+
+config OPTION_C
+	bool "C"
+	depends on MODE
+
+endchoice
+choice
+    int 365 if HOW_MUCH_DAYS_IN_YEAR || !WEEKS && DAYS
+    default BUILDER_BUILD if DROID
+	prompt "Choice 1" if ALA_MA_KOTA
```
kilroyd: OK, so we're not standardizing indentation of lines yet. That's OK for this commit where I'd like to get the basic infrastructure in place. Do we think that we can do that with what we have (as this is going to be the most common problem)

mwidera: Yes changing spaces -> tabs or tabs -> spaces would require deeper changes in lex.py and it's for sure achievable but I guess it would require changes in YACC parser also, which will be harder than this change I guess

kilroyd: I took a deeper look. Just for addressing this kind of indentation difference, I don't think we need to touch the lexer or grammar: we should be able to figure out what's happening in mconfigfmt.py and apply appropriate formatting directly there.

PR #65, opened 2019-02-26
Accept comparison operators in depends on

mwidera:
The existing code handling 'depends on' supports comparison operators,
but the menu indentation logic for dependent configs throws an exception
during init_config on encountering this construct.

Fix the menu indentation logic to ignore any 'depends on' with more
complicated expressions.

Add new test to check functionality.

Change-Id: I41d293b9620cd121d27bf37da4c966a4e3cc1d56

### DIFF
--- a/config_system/tests/depends_on_compare.test
+++ b/config_system/tests/depends_on_compare.test
@@ -1,0 +1,38 @@
+# This test checks 'depends on' statements using operators other than '&&' and '||'
+# Description of dependencies checks verification:
+# - Configuration which depends on some configurations and string
+# - Configuration which depends on integer value
+
+config STRING_OPTION
+    string "String to compare in a depends clause"
+    default "no"
+
+config INT_OPTION
+    int
+    default 21
+
+config OPTION
+    bool "An option in the menu"
+    default y
+
+config OPTION2
+    bool "An option in the menu"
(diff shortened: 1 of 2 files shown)


config_system/tests/depends_on_compare.test
```diff
@@ -0,0 +1,38 @@
+# This test checks the case where config dependency is defined by option of other type than configuration one
+# Description of dependencies checks verification:
+# - Configuration which depends on some configurations and sting
```
chrisdiamand: s/sting/string? :)

mwidera: :man_facepalming: Tanks!

config_system/tests/depends_on_compare.test
```diff
@@ -0,0 +1,38 @@
+# This test checks the case where config dependency is defined by option of other type than configuration one
```
chrisdiamand: What do you mean by "option of other type than configuration one"?

Maybe a better description would be: "This test checks 'depends on' statements using operators other than '&&' and '||'"?

The issue was that the operators weren't supported, not that the dependency was on a config option with a different type.

mwidera: Sure

PR #71, opened 2019-03-01
Improve config error messages

kilroyd:
Change-Id: I0dfecc77d857e744d10f765755778ec75c85a4d5

### DIFF
--- a/scripts/host_explore.py
+++ b/scripts/host_explore.py
@@ -44,7 +44,7 @@ def check_output(command, dir=None):
         output = subprocess.check_output(command, cwd=dir).strip()
         output = output.decode(sys.getdefaultencoding())
     except OSError as e:
-        logger.error(str(e))
+        logger.error("%s executing '%s'" % (str(e), command[0]))
     except subprocess.CalledProcessError as e:
         logger.warning("Problem executing command: %s" % str(e))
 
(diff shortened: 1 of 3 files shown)


scripts/host_explore.py
```diff
@@ -44,7 +44,7 @@ def check_output(command, dir=None):
         output = subprocess.check_output(command, cwd=dir).strip()
         output = output.decode(sys.getdefaultencoding())
     except OSError as e:
-        logger.error(str(e))
+        logger.error("%s executing '%s'" % (str(e), command))
```
chrisdiamand: `command` is an array usually - so this won't be very pretty, although it should still work.

You can do something like `" ".join([i for i in pipes.quote(i)])`? Although it is possible overkill :)

mwidera: Well if it's always a list then:
```suggestion
        logger.error("%s executing '%s'" % (str(e), ' '.join(command)))
```
if it can be string also, then it require one more check:
```suggestion
        logger.error("%s executing '%s'" % (str(e),'{}'.format(' ' if isinstance(x
command, list) else '').join(command)))
```

kilroyd: command is always a list, as this is what subprocess.check_output expects

Rather than import a separate module, I'm just printing command[0]. This should be enough to identify why we're getting an OSError. Problems encountered by the command should result in the CalledProcessError instead, and that should appropriately escape the command.

chrisdiamand: Sure, that works

PR #78, opened 2019-03-07
Update blueprint SHA

mwidera:
To avoid problems with latest release of golang 1.12
SHA updated to the newest commit

Change-Id: Ib048baf5a9f7d53b093d706afd6aa37589fae184


mwidera:
We cannot drop the support for golang 1.8 therefore for now this PR need to be closed

PR #86, opened 2019-03-15
Improve log messages visibility in console output

mwidera:
Add colored messages for console logging to provide more visible output
for the users.

Change-Id: Ifa22fc586fa4ae86ce3fd0fa7d08befa35001f43

### DIFF
--- a/config_system/log_handlers.py
+++ b/config_system/log_handlers.py
@@ -1,4 +1,4 @@
-# Copyright 2018 Arm Limited.
+# Copyright 2018-2019 Arm Limited.
 # SPDX-License-Identifier: Apache-2.0
 #
 # Licensed under the Apache License, Version 2.0 (the "License");
@@ -14,6 +14,7 @@
 # limitations under the License.
 
 import logging
+from copy import deepcopy
 from logging.handlers import BufferingHandler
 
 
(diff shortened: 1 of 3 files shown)


config_system/log_handlers.py
```diff
@@ -74,3 +75,28 @@ def errors(self):
...
+
+    def __init__(self, fmt, enabled=False):
+        super(ColorFormatter, self).__init__(fmt)
+        self.enabled = enabled
+
+    def format(self, log_msg):
+        paint = self.colors.get(log_msg.levelname)
+        if paint and self.enabled:
+            log_msg = deepcopy(log_msg)
+            log_msg.levelname = self.color_fmt.format(paint) + log_msg.levelname + self.reset
```
chrisdiamand: Because `colors`, `color_fmt` and `reset` aren't member variables, would it be more correct to refer to them as `ColorFormatter.colors` etc rather than via `self`?

mwidera: They are ColorFormatter class attributes, thus `self` is the proper way to access them.

kilroyd: Class variables (attributes) can be accessed either via self (`self.color_fmt`) or the class name (`ColorFormatter.color_fmt`).

If we use `self` for access it's not clear at the point we are using the variable that this is a class variable, and we could unintentionally re-bind the variable (by assigning to it).

If we use the class name for access then a sub-class of `ColorFormatter` could not override the variable.

I think it makes sense to access `color_fmt` and `reset` via the class names - a subclass should never have to override these.

That leaves `colors`. I suggest that we initialise `colors` as an instance variable in `__init__`. In the future we might wither want to subclass `ColorFormatter`, or provide functions to modify log coloring i.e. `setColor()`

mwidera: Such separation can be achieved with private class attribute.
Changed as requested.

PR #90, opened 2019-03-28
Replace string concatenation with string builder

mwidera:
Profiling of the Bob project shows that the utilization of the
string builder which has been introduced in Go 1.10 improves builder
execution time on Android

Change-Id: I023bb9bcfa94b902d9ec48089606bfceec0b7f67

### DIFF
--- a/core/android_make.go
+++ b/core/android_make.go
@@ -43,14 +43,15 @@ type androidMkGenerator struct {
 
 var androidLock sync.Mutex
 
-func writeIfChanged(filename string, text string) {
+func writeIfChanged(filename string, sb *strings.Builder) {
 	mustWrite := true
+	text := sb.String()
 
 	// If any errors occur trying to determine the state of the existing file,
 	// just write the new file
 	fileinfo, err := os.Stat(filename)
 	if err == nil {
-		if fileinfo.Size() == int64(len(text)) {
+		if fileinfo.Size() == int64(sb.Len()) {
 			current, err := ioutil.ReadFile(filename)
 			if err == nil {
 				if string(current) == text {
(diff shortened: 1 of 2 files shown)


core/android_make.go
```diff
@@ -72,9 +73,9 @@ func writeIfChanged(filename string, text string) {
 	}
 }
 
-func androidMkWriteString(ctx blueprint.ModuleContext, name string, text string) {
+func androidMkWriteString(sb *strings.Builder, ctx blueprint.ModuleContext, name string) {
```
chrisdiamand: Pref: Can `sb` be the last argument, as `text` was before?

That would follow the reasonably common convention of having the "output" (in this case, `name`, which is used to generate the output filename) earlier than the input (the `strings.Builder`).

It also makes it consistent with `writeIfChanged`.

core/android_make.go
```diff
@@ -667,89 +662,92 @@ func (g *androidMkGenerator) generateCommonActions(m *generateCommon, ctx bluepr
...
+		g.generateCommonActions(&sb, &m.generateCommon, ctx, inouts)
+		installGeneratedFiles(&sb, m, ctx, m.generateCommon.Properties.Tags)
+		androidMkWriteString(&sb, ctx, m.altShortName())
 	}
 }
 
 func (g *androidMkGenerator) genStaticActions(m *generateStaticLibrary, ctx blueprint.ModuleContext, inouts []inout) {
 	if enabledAndRequired(m) {
-		text := g.generateCommonActions(&m.generateCommon, ctx, inouts)
+		var sb strings.Builder
```
chrisdiamand: I wonder if you could declare it as `sb := &strings.Builder{}`?

That way, it will always be a pointer, so you can pass it to the functions below as `sb` (instead of `&sb`).

core/android_make.go
```diff
@@ -140,14 +141,14 @@ func specifyCompilerStandard(varname string, flags []string) string {
 	return line
 }
 
-func (m *library) GenerateBuildAction(binType int, ctx blueprint.ModuleContext) {
+func (m *library) GenerateBuildAction(sb *strings.Builder, binType int, ctx blueprint.ModuleContext) {
```
chrisdiamand: And in keeping with my comment on line 76, this function (and the other helpers, e.g. `declare*()`) are already correct in having `sb`, the "destination", as the first argument.

PR #91, opened 2019-03-28
Add source dir prefix to the ninja

mwidera:
This change adds source directory prefix to ensure that every ninja
build element have a proper path and doesn't relies on relative path.
To avoid colisions in path definition for Linux and Android
implementation source getting function has been split to relative
path and work directory path.

Change-Id: I7e808f70981ebe1e4909e03ee9694b5636cd7f6c

### DIFF
--- a/core/linux.go
+++ b/core/linux.go
@@ -69,15 +69,15 @@ func addPhony(p phonyInterface, ctx blueprint.ModuleContext,
 }
 
 func (g *linuxGenerator) sourcePrefix() string {
-	return "$SrcDir"
+	return "${SrcDir}"
 }
 
 func (g *linuxGenerator) buildDir() string {
-	return "$BuildDir"
+	return "${BuildDir}"
 }
 
 func (g *linuxGenerator) sourceOutputDir(m *generateCommon) string {
-	return filepath.Join("$BuildDir", "gen", m.Name())
+	return filepath.Join("${BuildDir}", "gen", m.Name())
 }
 
 var copyRule = pctx.StaticRule("copy",
@@ -99,13 +99,13 @@ type targetableModule interface {
 // Where to put generated shared libraries to simplify linking
 // As long as the module is targetable, we can infer the library path
 func getSharedLibLinkPath(t targetableModule) string {
-	return filepath.Join("$BuildDir", t.getTarget(), "shared", t.outputName()+".so")
+	return filepath.Join("${BuildDir}", t.getTarget(), "shared", t.outputName()+".so")
 }
 
 // Where to put generated binaries in order to make sure generated binaries
 // are available in the same directory as compiled binaries
 func getBinaryPath(t targetableModule) string {
-	return filepath.Join("$BuildDir", t.getTarget(), "executable", t.outputName())
+	return filepath.Join("${BuildDir}", t.getTarget(), "executable", t.outputName())
 }
 
 // Generate the build actions for a generateSource module and populates the outputs.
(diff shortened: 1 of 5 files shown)


core/linux.go
```diff
@@ -316,11 +316,13 @@ func (l *library) CompileObjs(ctx blueprint.ModuleContext) []string {
 		buildWrapper, buildWrapperDeps := l.Properties.Build.getBuildWrapperAndDeps(ctx)
 		args["build_wrapper"] = buildWrapper
 
-		sourceWithoutBuildDir := source
-		if buildDir := g.buildDir(); strings.HasPrefix(sourceWithoutBuildDir, buildDir) {
-			sourceWithoutBuildDir = sourceWithoutBuildDir[len(buildDir):]
+		output := l.ObjDir() + source + ".o"
+		if buildDir := g.buildDir(); strings.HasPrefix(source, buildDir) {
```
chrisdiamand: Pref: This is a bit hard to follow - you've got an `else` clause, but are also overwriting stuff in the `if` part.

Maybe it could be written as:
```go
var sourceWithoutPrefix string
if buildDir := g.buildDir(); strings.HasPrefix(source, buildDir) {
    sourceWithoutPrefix = source[len(buildDir):]
    // `source` already has a buildDir prefix
} else {
    sourceWithoutPrefix = source
    source = filepath.Join(g.sourcePrefix(), source)
}
var output = l.ObjDir() + sourceWithoutPrefix + ".o"
```
(approximately)

core/linux.go
```diff
@@ -755,6 +757,9 @@ func (*linuxGenerator) install(m interface{}, ctx blueprint.ModuleContext) []str
 
 	for _, src := range ins.filesToInstall(ctx) {
 		dest := filepath.Join(installPath, filepath.Base(src))
+		if !strings.HasPrefix(src, g.buildDir()) {
```
chrisdiamand: Is it possible to do this at the module type level?
- `bob_resource` always installs things from the source dir
- Every other module type always installs from the build dir

We should be able to do this reliably that way, rather than guessing based on the existing prefix

scripts/kmod_build.py
```diff
@@ -147,12 +147,16 @@ def build_module(output_dir, module_ko, kdir, module_dir, make_args, extra_cflag
...
     kconfig = os.path.join(abs_kdir, "linux", "kconfig.h")
-    for src_rel in args.sources:
-        if args.common_root:
-            dest = os.path.join(output_dir, os.path.relpath(src_rel, args.common_root))
-        else:
-            dest = os.path.join(output_dir, src_rel)
-        deps.extend(copy_with_deps.copy_with_deps(src_rel, dest, search_path, [kconfig]))
+    root = args.common_root
+    for src in args.sources:
+        if os.path.isabs(src):
```
chrisdiamand: I wonder if this check isn't actually correct - if `srcDir` is a relative path, we won't call `os.path.relpath`, and might actually get incorrect destination paths.

We need to ensure that every source file is under the common root, *even if the common root is relative*.

So possibly this should be more like:
```python
for src in args.sources:
    src = os.path.relpath(src, root)
    if src.startswith("../"):
        # Error
    dest = ...
    deps.extend(...)
```
Although I'm not sure of the most robust mechanism for ensuring that the sources are all subdirectories of the root - this thread has a few different suggestions: https://stackoverflow.com/questions/3812849/how-to-check-whether-a-directory-is-a-sub-directory-of-another-directory

core/linux.go
mwidera: Prefixing it at higher level is not possible due to the interference with the Android but it's possible to check type, so I've changed it. Also I've introduced the comment for further future, so it will be more obvious what's being done

scripts/kmod_build.py
mwidera: Python 3.5 implements `commonpath` [method](https://docs.python.org/3.7/library/os.path.html#os.path.commonpath) but, so far we need to support Python 2.7 also, thus suggested change is the right one. Changed

core/linux.go
mwidera: Ok sure if this is hard to follow then I've changed it to be more readable

PR #101, opened 2019-04-11
Support kernel build with clang for Android

mwidera:
Add configuration options allowing the user to control whether clang is
used for kernel module builds. Update kmod_build.py to pass the options
on to kernel the make call as required.

Change-Id: I2742e4562a6f66d7e0bf56a86733129549fb0dae

### DIFF
--- a/core/kernel_module.go
+++ b/core/kernel_module.go
@@ -118,6 +118,7 @@ func (m *kernelModule) generateKbuildArgs(ctx blueprint.ModuleContext) map[strin
 	var extraIncludePaths []string
 
 	g := getBackend(ctx)
+	props := getConfig(ctx).Properties
 
 	extraCflags := m.build().BuildProps.Cflags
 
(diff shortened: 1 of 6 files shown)


scripts/kmod_build.py
```diff
@@ -130,7 +136,7 @@ def build_module(output_dir, module_ko, kdir, module_dir, make_args, extra_cflag
     # Don't abspath cross_compile if it's just a prefix for something already
     # inside $PATH (i.e. it doesn't contain a directory part):
     cross_compile = args.cross_compile
-    if os.path.dirname(args.cross_compile):
+    if cross_compile and os.path.dirname(args.cross_compile):
```
kilroyd: We need the same logic for args.cc and args.hostcc

The caller could specify a relative path to the compiler to use, so kmod_build should make it absolute

core/kernel_module.go
```diff
@@ -139,6 +140,22 @@ func (m *kernelModule) generateKbuildArgs(ctx blueprint.ModuleContext) map[strin
 		kbuildOptions = "--kbuild-options " + strings.Join(m.build().Kbuild_options, " ")
 	}
 
+	hostToolchain := ""
+	if props.GetBool("host_toolchain_clang") {
+		hostToolchain = "--hostcc clang"
```
kilroyd: Here, "clang" should be HOST_CLANG_PREFIX + CLANG_CC_BINARY. Ideally we want to get this from toolchain

mwidera: I've changed this part to get tool from toolchain

scripts/kmod_build.py
mwidera: Done. I've introduced small function to avoid code duplication

tests/Mconfig
```diff
@@ -57,6 +57,22 @@ config TARGET_CLANG_TRIPLE
 	string
 	default "x86_64-linux-gnu"
 
+config KERNEL_CC
+	string
+	default "clang" if ANDROID
+	default "gcc"
```
kilroyd: Please remove the `default "gcc"` in example/Mconfig and tests/Mconfig. That way we leave the default compiler to the kernel

mwidera: Removed

PR #127, opened 2019-05-17
Support encapsulation of bob_generated_sources

mwidera:
The `encapsulates` property names modules that should be encapsulated by
the current bob_generate_source module. In this commit, just deal with
encapsulation of headers by re-exporting exported_gen_include_dirs.

Change-Id: I1114ce7ad3ace970c98b8d4e399abad226e87dba


core/library.go
```diff
@@ -409,7 +409,7 @@ func (l *library) GetExportedVariables(ctx blueprint.ModuleContext) (expLocalInc
 		if ctx.OtherModuleDependencyTag(dep) == wholeStaticDepTag ||
 			ctx.OtherModuleDependencyTag(dep) == staticDepTag ||
 			ctx.OtherModuleDependencyTag(dep) == sharedDepTag ||
-			ctx.OtherModuleDependencyTag(dep) == flagDepTag {
+			ctx.OtherModuleDependencyTag(dep) == reexportLibsTag {
```
chrisdiamand: OK, so in this commit you've also clarified/renamed the `reexport_libs` fields. Can you mention that in the commit message please?

docs/module_types/common_generate_module_properties.md
```diff
@@ -65,6 +65,12 @@ When `true` one job will run at a time - they won't be concurrent.
 Additional include paths to add for modules that use `generated_headers`. This
 will be defined relative to the module-specific build directory.
 
+----
+### **bob_generated.encapsulates** (optional)
+A list of source modules that this bob_generated_source will encapsulate.
+When this module is used with generated_headers, the named modules export_gen_include_dirs will be forwarded.
```
chrisdiamand: Grammar: "the named modules' `export_gen_include_dirs`..." (missing apostrophe)

docs/module_types/common_generate_module_properties.md
```diff
@@ -65,6 +65,12 @@ When `true` one job will run at a time - they won't be concurrent.
...
+When this module is used with generated_sources, the named modules outputs will be supplied as sources.
```
chrisdiamand: same here

tests/source_encapsulation/build.bp
```diff
@@ -0,0 +1,134 @@
...
+bob_binary {
+    name: "validate_source_encapsulation_simple",
+    generated_headers: ["encapsulation_source3"],
+    srcs: ["simple.c"],
+}
+
+bob_binary {
+    name: "validate_source_encapsulation_complex",
+    generated_headers: ["encapsulation_source3_complex"],
+    srcs: ["complex.c"],
```
chrisdiamand: OK, so this test case *should* get source 1, 2, 3 gen include dirs, as well as 4's, but *not* 5's. 

Could we test this by adding a `build_wrapper` script, which looks for `-I` flags in the command line and checks that none of them contain `source5`?

core/generated.go
```diff
@@ -616,7 +620,7 @@ func encapsulateMutator(mctx blueprint.TopDownMutatorContext) {
 	}
 
 	mctx.WalkDeps(func(child blueprint.Module, parent blueprint.Module) bool {
-		if mctx.OtherModuleDependencyTag(child) != generatedDepTag {
+		if mctx.OtherModuleDependencyTag(child) != encapsulationTag {
 			return false
 		}
 		parentProp, pok := getGenerateCommon(parent)
```
chrisdiamand: We should only have encapsulationTags on generated sources, because we only added them to the correct module types.

This means we can remove the checks on line 630 and 633.

core/generated.go
```diff
@@ -34,6 +34,7 @@ var (
 	generatedHeaderTag = dependencyTag{name: "generated_headers"}
 	generatedSourceTag = dependencyTag{name: "generated_sources"}
 	generatedDepTag    = dependencyTag{name: "generated_dep"}
+	encapsulationTag   = dependencyTag{name: "source_encapsulation"}
```
chrisdiamand: pref: `encapsulatesTag` for consistency

core/generated.go
chrisdiamand: (Not sure if github's showing this correctly, but this comment only applies to the second commit!)

docs/module_types/common_generate_module_properties.md
mwidera: Sure, but you probably mean "module's"

chrisdiamand: Nope - there can be multiple modules, so the apostrophe goes after the 's' :)

core/library.go
chrisdiamand: I think this is still not done?

PR #130, opened 2019-05-24
Use Rpath $ORIGIN for Linux executables

mwidera:
For Linux related executables and shared libraries gather shared library
deps paths and recalculate relative path from all of it (including given executable)
The outcome of it prefix with $ORIGIN
This is interpreted by the linker so all dependent shared libraries will
be included with -Rpath token as described in ld.so documentation

Change-Id: Ic557d1a514bbeff5989068ba7c6163d0cd77fec4

### DIFF
--- a/core/install.go
+++ b/core/install.go
@@ -87,6 +87,15 @@ type InstallableProps struct {
 	Post_install_tool string
 	// Command to execute on file(s) after they are installed
 	Post_install_cmd string
+	// The path retrieved from the install group so we don't need to walk dependencies to get it
+	Install_path *string `blueprint:"mutated"`
+}
+
+func (props *InstallableProps) getInstallGroupPath() (string, bool) {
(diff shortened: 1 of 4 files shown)


core/install.go
```diff
@@ -272,3 +250,33 @@ func resourceFactory(config *bobConfig) (blueprint.Module, []interface{}) {
...
+			if !ok {
+				panic(fmt.Sprintf("install_group dependency of %s not an install group", mctx.ModuleName()))
+			}
+			if installGroupPath != nil {
+				panic(fmt.Sprintf("Multiple install group dependencies for %s", mctx.ModuleName()))
+			}
+			installGroupPath = &insg.Properties.Install_path
+		})
+	if installGroupPath == nil {
+		return
```
kilroyd: Avoid the return here please, and nest the remaining code

PR #132, opened 2019-06-03
Allow usage of implicit output for generated sources

mwidera:
Add option to generate source which will implicitly define output
This change is mostly related to the ninja output definition
Add function to retrieve implicit output for some of the types and also
define it on the dependent interface. Depfiles now are including also
implicit output, so e.g. headers might be passed through
Now all of them follow implicit naming scheme.
This change also aligns documentation for the generated/transform source
sections

Lastly added test to verify modified content of implicit output handling

Change-Id: I62233521ddc7b7b5058b9c45cbee62cef6f27c8d


core/android_make.go
```diff
@@ -648,13 +648,14 @@ func (g *androidMkGenerator) generateCommonActions(sb *strings.Builder, m *gener
 	for _, inout := range inouts {
 		if _, ok := args["headers_generated"]; ok {
 			headers := utils.Filter(inout.out, utils.IsHeader)
-			args["header_generated"] = strings.Join(headers, " ")
+			implHeaders := utils.Filter(inout.implicitOuts, utils.IsHeader)
+			args["header_generated"] = strings.Join(append(headers, implHeaders...), " ")
```
chrisdiamand: I think you can use `utils.Join` here?

core/build_structs.go
```diff
@@ -67,6 +67,7 @@ type TargetSpecific struct {
 type dependentInterface interface {
 	phonyInterface
 	outputDir(g generatorBackend) string
+	implicitOutputs(g generatorBackend) []string
```
chrisdiamand: Would this not make more sense in `phonyInterface` next to `outputs()`? It seems a bit strange to have different kinds of outputs in different interfaces.

(Admittedly I think both should be in `dependentInterface`, but we shouldn't change that here).

kilroyd: I agree that it looks odd. However, I don't think implicitOutputs belongs in phonyInterface - phony support only requires outputs()

kilroyd: Thinking about this again, it's possible to define a module that only has implicit outputs (no explicit outputs). A phony rule must therefore depend on both the explicit and implicit outputs.

So please add implicitOutputs() to the phonyInterface, and update phony target generation to include implicitOutputs(). Eventually we might want an allOutputs() function.

PR #136, opened 2019-06-11
Noop soong plugin

chrisdiamand:
Create basic bootstrap scripts and a soong plugin which adds the module types. This allows the `Android.bp` files (which are links to `build.bp` files created at bootstrap time) to be parsed, although no build actions will be generated.

Currently, this requires manually generating a `config.json` file and copying it into e.g. `out/target/product/hikey960/gen/STATIC_LIBRARIES/bob-tests-config/config.json`.


core/soong_plugin.go
```diff
@@ -0,0 +1,117 @@
+// +build ignore
```
kilroyd: Can we put this after the copyright declaration please. I believe it just needs to be before the `package` declaration. You might even put it after the comment describing why we have it.

kilroyd: I assume microfactory is ignoring the build tags and doing exactly what the Android.bp says. So can we use `+build soong` here? That way we could do some testing (by specifying --tags soong)

chrisdiamand: I tried putting it after the copyright and comment first, unfortunately it doesn't work. Also no luck just after the copyright. In both cases I've tried various amounts of surrounding whitespace.

Microfactory is indeed ignoring the tags, since it's overridden by `Android.bp`.

`+build soong` does work, and it means I can cause e.g. `go vet` to re-include the file with `go vet --tags soong github.com/ARM-software/bob-build/core`.

PR #143, opened 2019-06-20
Support using specific files from sources

mwidera:
Support {{match_source file}} matching in string properties
to reference a subset of explicit sources.

Change-Id: I300c1a3d72b455d7ac438075a33de690d9c57e03


tests/match_source/build.bp
```diff
@@ -0,0 +1,44 @@
...
+}
+
+bob_generate_source {
+    name: "match_source_gen",
+    srcs: [
+        "main.c",
+        "custom.txt",
+    ],
+    out: ["custom_src_gen"],
+    cmd: "cat {{match_srcs \"custom.txt\"}} {{match_srcs \"main.c\"}} > $out",
```
chrisdiamand: This would still pass if `{{match_srcs}}` returned e.g. an empty string, or e.g. matched the same source file twice.

Maybe we should write some C code in `custom.txt` too, then compile the generated code in another `bob_binary` which depends on the code in `main.c` as well as `custom.txt`.

utils/utils.go
```diff
@@ -29,16 +30,28 @@ var (
 	headerRegexp = regexp.MustCompile(`\.(h|hpp|inc)$`)
 )
 
+var compileSrcsExt = []string{".c", ".s", ".cpp", ".cc", ".S"}
```
chrisdiamand: Fix: Isn't this unused?

chrisdiamand: Ah sorry, this is used.

Can we implement this by matching the pattern of `IsHeader`/`IsNotHeader`? So make this a regex like `headerRegexp`?

mwidera: Sure changed to regex

core/late_template.go
```diff
@@ -0,0 +1,111 @@
...
+
+func (s *SourceProps) matchSources(ctx blueprint.BaseModuleContext, arg string) string {
+	g := getBackend(ctx)
+
+	for _, match := range matchSourcesRegex.FindAllStringSubmatch(arg, -1) {
+		matchFlag := false
+		if len(match) <= 0 {
+			panic("Invalid argument for match_srcs. match_srcs expects a single pattern used to match a file.")
+		}
+		params := []string{}
```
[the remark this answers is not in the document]
mwidera: Ok name changed to indicate better purpose of it

core/late_template.go
```diff
@@ -0,0 +1,111 @@
...
+		for _, src := range s.getSources(ctx) {
+			matched, err := filepath.Match("*/"+match[1], src)
+			if err != nil {
+				panic("Error during matching filepath pattern")
+			}
+			if matched {
+				params = append(params, filepath.Join(g.sourcePrefix(), src))
+				matchFlag = true
```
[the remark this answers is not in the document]
mwidera: Yeah changed

tests/match_source/build.bp
mwidera: Yeah that's a good idea and I believe this should be made as a follow-up change

PR #144, opened 2019-06-21
Add pylint checks to travis

mwidera:
Also fix couple of errors for both python versions

Change-Id: Id274d4b4378e015fd2d8796a753c31e8204d4b3f

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -26,7 +26,8 @@ before_install:
     sudo apt-get install -y rsync python3 \
      python-ply  python-pytest  python-pytest-catchlog  python-mock  python-pytest-mock \
     python3-ply python3-pytest python3-pytest-catchlog python3-mock python3-pytest-mock \
-    python-pep8 python3-pep8
+     python-pep8 pylint \
+    python3-pep8 pylint3
 
 install:
   - bash .travis/install-ninja.sh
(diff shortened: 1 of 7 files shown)


.travis.yml
```diff
@@ -24,9 +24,10 @@ before_install:
   - sudo apt-get update
   - |
     sudo apt-get install -y rsync python3 \
-     python-ply  python-pytest  python-pytest-catchlog  python-mock  python-pytest-mock \
+    python-ply  python-pytest  python-pytest-catchlog  python-mock  python-pytest-mock \
```
kilroyd: I think Chris was aligning the lines on the '-' so we can see we have the same things installed for python2 and python3. Can you keep to that please?

mwidera: Ok changed, so requirements are now in aligned

kilroyd:
The test failure looks to be due to a change in the way travis checks out the PR commit. Please ignore the bootstrap test failure which we'll fix separately

config_system/config_system/general.py
```diff
@@ -756,11 +756,7 @@ def get_default_style():
 
 
 class StyledText(object):
-    def __init__(self):
-        self.style = get_default_style()
-        self.text = ""
-
-    def __init__(self, text):
+    def __init__(self, text=""):
```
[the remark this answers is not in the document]
mwidera: It is just a merge of duplicated constructors for this class. After consultation default value is not needed anymore, thus it was removed (see latest push)

PR #150, opened 2019-07-01
Standardise and document template functions

kilroyd:
We have a few different custom functions available via Go templates,
with varying naming convention. Standardise on snake case. Also prefix
regular expression functions with `reg_`.

Since we support `to_upper`, also support the converse, `to_lower`.

With {{match_srcs}} reduce the number of properties we allow it on.
The issue is that depending on the property, the dependencies need to
be setup slightly differently. Things kind-of work as is, but there
could be some subtle issues. Therefore support:

 * the primary case where ldflags needs to reference a file

 * ability for bob_generate_* to reference specific files in its cmd

Change-Id: I793933dc4e67d3fcea25c958a76dbbb4eb5fd25e

### DIFF
--- a/core/late_template.go
+++ b/core/late_template.go
@@ -54,20 +54,18 @@ func (s *SourceProps) matchSources(ctx blueprint.BaseModuleContext, arg string) 
 	return arg
 }
 
-// This mutator applies match result for required glob or filename from sources list.
-// It searches through build properties:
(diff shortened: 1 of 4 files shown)


kilroyd:
The bootstrap failure on the pull request test looks like a false positive, and an issue that we've seen before. I'm merging this anyway...

PR #151, opened 2019-07-01
Fix minor lint issues

kilroyd:
Use a consistent identifier for externalLib.

Remove an unused 2nd loop argument.

When setting up the workspace, ensure testify is available.

Change-Id: Iadddf2a07f4559afbd0ddc901a5ddd6357482b14

### DIFF
--- a/core/external_library.go
+++ b/core/external_library.go
@@ -26,7 +26,7 @@ type externalLib struct {
 	blueprint.SimpleName
 }
 
-func (l *externalLib) topLevelProperties() []interface{} { return []interface{}{} }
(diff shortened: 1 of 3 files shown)


kilroyd:
Pull request re-bootstrap check is failing again, I don't think it should be a problem

chrisdiamand:
> Pull request re-bootstrap check is failing again, I don't think it should be a problem

Yep - the copyright year checker is also failing. I think we need to work out how we should be checking commit ranges for pull requests.

The branch checks are fine though so I'm not too worried.

PR #158, opened 2019-07-11
Travis test rework

kilroyd:
Miscellaneous cleanup to avoid some spurious error we were seeing, and simplifications.


.travis/set_python_version.sh
```diff
@@ -14,16 +15,20 @@ if [[ -n $PYTHON_SUFFIX ]]; then
...
+        PYTHON_ERR=1
     fi
 else
     echo "Error: No Python version selected"
-    return 1
+    PYTHON_ERR=1
 fi
 
-return 0
+# If we had a problem above, run false to set the exit code so the
```
chrisdiamand: What exactly was going wrong here, vs using `return`?

kilroyd: This was me seeing a few different errors in Travis, twiddling a few things to check and cleaning up.

I'll re-instate the returns, as they should work and it's a bit cleaner (though I'm not keen on a. source'd scripts returning b. source'd scripts using exit codes)

chrisdiamand: Thanks! Yeah I guess ideally they should be in a function.

PR #159, opened 2019-07-15
Make menuconfig arguments available to plugins

kilroyd:
In commit e53161f277b we modified the prototype of init_config() to
record the filename that would be used to save the config file, for
use by a plugin. This broke some external code that is also calling
init_config().

It doesn't actually make sense to pass a filename in this call, so
come up with an alternative way to get the filename.

Provide a wrapper for parse_args(), so that we can keep a copy of all
the arguments for any invocation. Provide wrapper functions to get the
appropriate options. Remove some custom code that was doing something
similar for the database directory.

Change-Id: I0dfb052986581a4aff61d948554607737d49af68

### DIFF
--- a/config_system/config_system/utils.py
+++ b/config_system/config_system/utils.py
@@ -26,6 +26,9 @@ except ImportError:
 logger = logging.getLogger(__name__)
 
 
+__args = None
+
+
 @contextlib.contextmanager
 def open_and_write_if_changed(fname):
     """Return a file-like object which buffers whatever is written to it. When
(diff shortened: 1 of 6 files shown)


config_system/config_system/utils.py
```diff
@@ -47,3 +50,27 @@ def open_and_write_if_changed(fname):
             logger.debug("Updating {}".format(fname))
             with open(fname, "wt") as fp:
                 fp.write(buf.getvalue())
+
+
+def get_config_dir():
```
chrisdiamand: I wonder if we want to have these return `__args.config` unchanged? (i.e. without running `dirname`)

This would allow more flexibility in the callers - it shouldn't be a problem for them to run `os.path.dirname` themselves, and it'll be easy to get the actual filenames then if we need them later.

kilroyd: I think having functions identify what the callers actually want is useful, and separates what the plugins want to know about from how the argument(s) are passed in / structured.

If something wanted to know the config filename, I would say we add a get_config_file() wrapper.

kilroyd:
Discussing this with Chris:
* Mconfig filename should be tied to the arg passed to init_config() (and therefore 'stored' in general). This is always available.
* config directory is only available once you read_config() or write_config(), so needs to come from 'somewhere else' if a plugin is using it. Otherise it should also be 'stored' in general. Keeping the args around like this commit does achieves this, but feels like a workaround.

I'm look at pushing some alternative approaches, and we can choose one of them...

PR #162, opened 2019-07-16
Move JSON output to main update_config

kilroyd:
And update the format to be a simple list of properties. Make the go code construct the feature lists that it needs from the properties.

Also separate out the environment hashing into a Bob script.

### DIFF
--- a/bob.bash
+++ b/bob.bash
@@ -43,6 +43,10 @@ python "${BOB_DIR}/config_system/update_config.py" \
        --bob-config "${BUILDDIR}/config.json" \
        ${BOB_CONFIG_OPTS}
 
+# Get a hash of the environment so we can detect if we need to
+# regenerate the build.ninja
+python "${BOB_DIR}/scripts/env_hash.py" "${BUILDDIR}/.env.hash"
+
 # Source the pathtools script - we need bob_realpath for CCACHE_BASEDIR.
 source "${BOB_DIR}/pathtools.bash"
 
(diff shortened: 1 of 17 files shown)


bob.bash
```diff
@@ -43,6 +43,10 @@ python "${BOB_DIR}/config_system/update_config.py" \
        --bob-config "${BUILDDIR}/config.json" \
        ${BOB_CONFIG_OPTS}
 
+# Get a hash of the environment so we can detect if we need to
+# regenerate the build.ninja
+python "${BOB_DIR}/scripts/env_hash.py" "${BUILDDIR}/.env.hash"
```
chrisdiamand: Probably for a follow-up commit, but we ought to consider whether we need to do this in the config system wrappers too, as Soong might need this (which doesn't run `buildme`/`bob.bash`).

kilroyd: I think we're ok for soong. Android does an environment check and I believe it regenerates build.ninja if it sees it change. For us it would mean the bob modules are also reprocessed. See soong/cmd/soong_env/soong_env.go

utils/utils.go
```diff
@@ -261,3 +262,10 @@ func NewStringSlice(lists ...[]string) []string {
 	}
 	return allLists
 }
+
+// Exit the program, printing an message to stderr.
```
kilroyd: s/an/a/

bob.bash
kilroyd: OK, in order for that to work for us, our plugin needs to call android.Getenv(), and android will record the dependency on that environment variable.

PR #163, opened 2019-07-17
Module rename

chrisdiamand:
Some preparatory work to enable running mutators before we create the Soong modules.


chrisdiamand:
Only the `android.TopDownMutatorContext` has the `CreateModule` method, so we have to use top down for the load hook replacement.

I think we can do `Rename` on either - I pretty much chose at random for that one :)

PR #171, opened 2019-08-05
config_system: support '+' and '-' operators

kilroyd:
This commit series enables the use of '+' and '-' in expressions in Mconfig. For integers, this results in addition and subtraction. For strings, '+' results in concatenation.

This will simplify some configuration construction, where the default value of an option can be expressed as the combination of other options.

Note that these operators do not apply to booleans, where the 'default if' and 'select if' constructions provide this capability.

The initial commit implements the necessary support. The subsequent commits refactor the code to improve maintainability, and add testing.

### DIFF
--- a/config_system/config_system/general.py
+++ b/config_system/config_system/general.py
@@ -17,7 +17,7 @@ import logging
 import os
 import re
 
-from config_system import utils
+from config_system import expr, utils
 
 
 logger = logging.getLogger(__name__)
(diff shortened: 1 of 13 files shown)


config_system/config_system/general.py
```diff
@@ -43,105 +43,176 @@ def check_depends(depends, value):
...
+    """
+    assert type(e) == tuple
+    assert len(e) in [2, 3]
+    if len(e) == 3:
+        left = _expr_value(e[1])
+        right = _expr_value(e[2])
+        if type(left) != type(right):
+            raise TypeError("'{}' operator is not valid with mixed types".format(e[0]))
+        elif left in ['y', 'n'] or right in ['y', 'n']:
+            raise TypeError("'{}' operator is not valid on booleans".format(e[0]))
```
chrisdiamand: Won't this incorrectly error if we wanted to append _strings_ where one of the operands is 'y' or 'n'?

I guess we could leave this commented out until we can convert booleans to using actual `bool` types.

Or actually, could `_expr_value()` return a tuple of `(type, value)` instead? Then we discard the type in the wrapper `expr_value`?

kilroyd: Yes, this will behave incorrectly if you want a string to have the value 'y' or 'n'. I don't believe we have instances of this in user configs, so we should be OK for now. We should just go straight to handling the booleans as a distinct type. We can do this in a follow up commit, as it will affect a fair amount of unrelated code.

chrisdiamand: OK - I've seen a nearly-ready version of a patch to go directly to actual bools, so this is bearable for now.

PR #172, opened 2019-08-14
config_system: use booleans for 'y' and 'n'

kilroyd:
Internally use the appropriate python datatype to handle booleans.
This simplifies type checking, and avoids needing to pass the expected
type around.

Change-Id: I25a3b7c9ee76a38f6bb2697f17e061c7cb3c2f6e

### DIFF
--- a/config_system/config_system/general.py
+++ b/config_system/config_system/general.py
@@ -25,14 +25,14 @@ logger.addHandler(logging.NullHandler())
 
 def can_enable(depends):
     value = expr.condexpr_value(depends)
-    if value == 'y':
+    if value is True:
         return True
     return False
 
 
 def is_visible(cond):
     value = expr.condexpr_value(cond)
-    if value == 'n':
+    if value is False:
         return False
     return True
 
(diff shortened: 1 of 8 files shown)


config_system/config_system/general.py
```diff
@@ -25,7 +25,7 @@
 
 def can_enable(depends):
     value = expr.condexpr_value(depends)
-    if value == 'y':
+    if value is True:
```
kilroyd: Missed an equivalent change in the is_visible() function just below.

Looking at these functions though, it's a bit odd that we take the condition as an argument instead of the config

chrisdiamand: Maybe we could simplify these two functions to just return `value` directly? (and maybe error if the type is wrong)

I don't mind if you do that here or in a follow up though

kilroyd: Yeah. I've a feeling that we rely on the explicit check. For example where condexpr_value returns the value of a string identifier. So this is definitely a follow-up candidate.

PR #186, opened 2019-09-02
Add test for library stripping

kilroyd:
... and fix the implementation after a last minute refactoring broke
an interface.

The android make implementation also needs fixing. Since
GenerateBuildAction act on the type library, it is always getting the
library implementation of strip() that returns false. Refactor so that
the function takes an interface. Keep the changes in here minimal for
now.

Change-Id: Ice5747bcd4a7919466ee1e47a08d9bc54baadd53

### DIFF
--- a/core/android_make.go
+++ b/core/android_make.go
@@ -143,7 +143,27 @@ func specifyCompilerStandard(varname string, flags []string) string {
 	return line
 }
 
-func (m *library) GenerateBuildAction(sb *strings.Builder, bt binType, ctx blueprint.ModuleContext) {
+// This function generates the Android make fragment to build static
+// libraries, shared libraries and executables. It's evolved over time
+// and needs to be refactored to use interfaces better.
+func androidLibraryBuildAction(sb *strings.Builder, mod blueprint.Module, ctx blueprint.ModuleContext) {
+	var bt binType
+	var m library
+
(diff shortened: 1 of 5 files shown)


core/android_make.go
```diff
@@ -271,7 +291,7 @@ func (m *library) GenerateBuildAction(sb *strings.Builder, bt binType, ctx bluep
 		sb.WriteString("LOCAL_MODULE_OWNER := " + m.Properties.Owner + "\n")
 		sb.WriteString("LOCAL_PROPRIETARY_MODULE := true\n")
 	}
-	if m.strip() {
+	if strlib, ok := mod.(stripable); ok && strlib.strip() {
```
chrisdiamand: Don't we expect all libraries to be stripable? We know this is definitely a binary, shared lib or static lib from the switch at the start of the function.

If we write this like:
```
if mod.(stripable).strip() {
}
```
Then we'll get a panic if we accidentally break libraries implementing `stripable`. At the moment, they would silently stop being stripped :)

kilroyd: At the moment we do, but that's not necessarily the case. Static libraries don't have to implement stripable. They do at the moment (and return false) as I was having problems making it work for both android and linux. I think I can now remove library.strip()

PR #188, opened 2019-09-04
Toolchain

kilroyd:
Various toolchain tweaks in aid of supporting different configurations of clang


core/toolchain.go
```diff
@@ -255,8 +260,7 @@ func newToolchainGnuNative(config *bobConfig) (tc toolchainGnuNative) {
 }
 func newToolchainGnuCross(config *bobConfig) (tc toolchainGnuCross) {
```
chrisdiamand: Looks like the cross and native versions are pretty much identical. At some point we might want to merge them, although let's leave it as-is for now.

kilroyd: Yes, the difference in the clang function should also be getting smaller, so eventually I expect we can eliminate the Cross versions

PR #190, opened 2019-09-09
Improve handling of clang configurations

kilroyd:
Various changes that allow users to tell Bob how their clang is configured, and therefore how to call it.

### DIFF
--- a/core/toolchain.go
+++ b/core/toolchain.go
@@ -278,7 +278,7 @@ type toolchainClangCommon struct {
 	clangBinary    string
 	clangxxBinary  string
 	prefix         string
-	useGnuLibs     bool
+	useLibgcc      bool
 	useGnuStl      bool
 	useGnuBinutils bool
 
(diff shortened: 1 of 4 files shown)


core/toolchain.go
```diff
@@ -80,26 +80,40 @@ func getToolPath(toolUnqualified string) (toolPath string) {
...
 	if err != nil {
-		fmt.Printf("Error: Couldn't get path for %s: %v\n", basename, err)
+		e = fmt.Errorf("Couldn't get path for %s: %v", basename, err)
 		return
 	}
 
-	fname := strings.TrimSpace(string(bytes))
+	fname = strings.TrimSpace(string(bytes))
+
 	if _, err := os.Stat(fname); os.IsNotExist(err) {
```
chrisdiamand: Did we want to expand this condition, e.g. general "cannot stat" errors, or if it's a directory/otherwise unreadable?

chrisdiamand: D'oh this is existing code. I'm getting mixed up with the compiler runtime commit.

kilroyd:
After some testing, we don't need _CLANG_USE_GNU_SYSTEM to set the defaults for _STL_LIBRARY and _COMPILER_RUNTIME. Having it is also forcing us to check what the default runtime library is wher GNU_SYSTEM does get used. Therefore remove _CLANG_USE_GNU_SYSTEM and drop the final commit that avoids setting --rtlib if it matches clangs default.

PR #199, opened 2019-09-24
Prevent build on configuration failure without removing config files

a-khabarov:
'bob.config' is no longer removed on configuration failure.
Instead, 'bob.error' is created to signal 'bob.bash' that the build
should not proceed furhter.

Change-Id: Idd2f8c9abd4e06e3615e9f6ad55e83fa556ae44e

### DIFF
--- a/config_system/menuconfig.py
+++ b/config_system/menuconfig.py
@@ -688,10 +688,21 @@ def main():
 
     error_count = counter.errors() + counter.criticals()
 
-    if should_save and error_count > 0:
-        os.remove(args.config)
-        if args.json is not None:
-            os.remove(args.json)
+    if should_save:
+        error_path = args.config + ".error"
+        if error_count == 0:
+            try:
+                os.remove(error_path)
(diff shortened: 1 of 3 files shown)


config_system/menuconfig.py
```diff
@@ -688,10 +689,17 @@ def main():
...
-            os.remove(args.json)
+    if should_save:
+        error_path = args.config + ".error"
+        if error_count == 0:
+            try:
+                os.remove(error_path)
+            except OSError:
+                pass
+        else:
+            subprocess.call(["touch", error_path])
```
kilroyd: Rather than use subprocess to call touch, please just call open

config_system/menuconfig.py
```diff
@@ -688,10 +689,17 @@ def main():
...
+            if args.json is not None:
+                os.remove(args.json)
```
kilroyd: catch the OSError here too please

PR #205, opened 2019-10-09
Changes depfile property to bool

micprz:
Change-Id: Icbc467aa2c3de54fbc6fee64ead2106f74824060


core/android_make.go
```diff
@@ -692,6 +693,11 @@ func (g *androidMkGenerator) generateCommonActions(sb *strings.Builder, m *gener
 	sb.WriteString("\n")
 
 	cmd, args, implicits, _ := m.getArgs(ctx)
+
+	if proptools.Bool(m.Properties.Depfile) && !strings.Contains(cmd, "${depfile}") {
+		panic(errors.New(fmt.Sprintf("%s depfile is true, but ${depfile} not used in cmd", m.Name())))
```
chrisdiamand: You can use `fmt.Errorf ` here to remove one layer of function calls.

core/generated.go
```diff
@@ -116,6 +118,9 @@ type GenerateProps struct {
 
 	// The target type - must be either "host" or "target"
 	Target tgtType
+
+	// If true, depfile name will be generated and can be used as $depfile reference in 'cmd'
```
chrisdiamand: Can you change `$depfile` to `${depfile}` in this comment please? That's what we check for in the error message above.

core/android_make.go
chrisdiamand: Also could we move it into `getArgs`, so that it's not duplicated between android_make and Linux?

core/soong_gen.go
```diff
@@ -274,7 +274,7 @@ func (gc *generateCommon) getHostBinModule(mctx android.TopDownMutatorContext) (
 }
 
 func (gc *generateCommon) createGenrule(mctx android.TopDownMutatorContext,
-	out []string, depfile string) {
+	out []string, depfile *bool) {
```
chrisdiamand: Could this be a simple `bool` type rather than `*bool`? We aren't using this value as a module property, so we don't need to use the obfuscated types required for properties :)

docs/module_types/bob_transform_source.md
```diff
@@ -31,9 +31,9 @@ bob_transform_source {
     out: {
         match: "file_([0-9])+.cpp",
         replace: ["new_$1.o"],
-        depfile: "$1.d",
         implicit_srcs: ["my_file.scu"],
     },
+    depfile: true,
```
chrisdiamand: Can we add a `${depfile}` reference to `cmd` in this documentation? We require that if depfile is true so it would be useful to mention it.

core/soong_gen.go
```diff
@@ -294,7 +294,7 @@ func (gc *generateCommon) createGenrule(mctx android.TopDownMutatorContext,
 		Tool:                    gc.Properties.Tool,
 		HostBin:                 hostBinModuleName,
 		Cmd:                     cmd,
-		Depfile:                 depfile,
+		Depfile:                 proptools.Bool(depfile),
```
micprz: I noticed this is incorrect too, string != bool

micprz: Changed Depfile type to bool as agreed with Chris.

PR #213, opened 2019-10-21
Abstract linker interface

auselen:
Abstract linker interface for better argument handling.

Change-Id: I6f437cca7d00564f33e23a9841d7858ac55f6bd2
Signed-off-by: Ali Utku Selen <ali.utku.selen@arm.com>

### DIFF
--- a/core/android_make.go
+++ b/core/android_make.go
@@ -153,7 +153,7 @@ func specifyCompilerStandard(varname string, flags []string) string {
 // This function generates the Android make fragment to build static
 // libraries, shared libraries and executables. It's evolved over time
 // and needs to be refactored to use interfaces better.
-func androidLibraryBuildAction(sb *strings.Builder, mod blueprint.Module, ctx blueprint.ModuleContext) {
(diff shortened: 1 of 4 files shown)


core/android_make.go
```diff
@@ -153,7 +153,7 @@ func specifyCompilerStandard(varname string, flags []string) string {
 // This function generates the Android make fragment to build static
 // libraries, shared libraries and executables. It's evolved over time
 // and needs to be refactored to use interfaces better.
-func androidLibraryBuildAction(sb *strings.Builder, mod blueprint.Module, ctx blueprint.ModuleContext) {
+func androidLibraryBuildAction(tcs toolchainSet, sb *strings.Builder, mod blueprint.Module, ctx blueprint.ModuleContext) {
```
kilroyd: Please move the toolchainSet to the end of the argument list.

The other arguments are the primary things that we are operating on, and the toolchain is secondary to what the function is doing.

auselen: ok.

PR #216, opened 2019-10-23
Add implicit dependencies on Module.symvers for soong

micprz:
Also:
- enhanced kernel module tests
- added implicit dependency on kmod_build
- disabled broken test "transform_source", see changeset:
09fa1251de7742f4bdaf2f20380661724bd5af62

Change-Id: I04cd4791c7453979c92b900b277bb789d4a6415b


core/soong_kernel_module.go
```diff
@@ -78,16 +90,42 @@ var soongKbuildRule = apctx.StaticRule("bobKbuild",
 	}, "kmod_build", "depfile", "extra_includes", "extra_cflags", "kbuild_extra_symbols", "kernel_dir", "kernel_cross_compile",
 	"kbuild_options", "make_args", "output_module_dir", "cc_flag", "hostcc_flag", "clang_triple_flag")
 
+
 func (m *kernelModuleBackend) GenerateAndroidBuildActions(ctx android.ModuleContext) {
 	builtModule := android.PathForModuleOut(ctx, m.Name()+".ko")
 	symvers := android.PathForModuleOut(ctx, "Module.symvers")
 
+	// preserve symvers location for this module (for the parent pass)
+	m.Symvers = symvers
```
kilroyd: OK, so as we process the build actions for each module we save the path to Module.symvers, and pick up the path in the walk. So we're relying on the order of the GenerateAndroidBuildAction calls to be be in dependency order (which I believe is true).

This doesn't feel like the proper way to do this, but is OK for now.

core/soong_kernel_module.go
```diff
@@ -78,16 +90,42 @@ var soongKbuildRule = apctx.StaticRule("bobKbuild",
...
+		func(m blueprint.Module) bool { return ctx.OtherModuleDependencyTag(m) == kernelModuleDepTag },
+		func(m blueprint.Module) {
+			if km, ok := m.(*kernelModuleBackend); ok {
+				depSymvers = append(depSymvers, km.Symvers)
+			} else {
+				panic(fmt.Errorf("%s not a kernel module backend", m.Name()))
+			}
+		})
+
+	// overwrite incorrect paths
```
kilroyd: Is this comment right? I don't think the paths are incorrect at this point - they're not there. Suggest just removing this comment

micprz: Comment is right, I moved it few lines below.

core/soong_kernel_module.go
micprz: Ok.

PR #230, opened 2019-11-13
Use getopts instead of getopt in Bob tests

a-khabarov:
When running Bob tests on OSX everything works correctly until
`tests/build_tests.sh` calls `tests/bootstrap.sh -o build-in-src`.

`tests/bootstrap.sh` invokes `getopt` to get the `BUILDDIR` argument
(that is supposed to be `build-in-src`), but this does not work and it defaults
to `BUILDDIR=build`, breaking everything that follows.

This is because `getopt` on OSX 10.13.3 is different from `getopt` on Linux.

This commit fixes this by switching to `getopts` which is dependent on the shell
and not on the OS.

Change-Id: I5035d30cfe8fcf05036bd5716875a732ad995858

### DIFF
--- a/tests/bootstrap
+++ b/tests/bootstrap
@@ -50,32 +50,22 @@ EOF
 
 source "${SCRIPT_DIR}/bootstrap_utils.sh"
 
-PARAMS=$(getopt -o "o:h" --name ${BASENAME} -- "$@")
-
-eval set -- "$PARAMS"
-unset PARAMS
-
-while true ; do
-    case $1 in
-        -o)
-            BUILDDIR=$2
-            shift 2
+while getopts "o:h" opt ; do
+    case "${opt}" in
+        o)
+            BUILDDIR="${OPTARG}"
             ;;
-        -h)
+        h)
             usage
             exit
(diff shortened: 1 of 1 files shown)


tests/bootstrap
```diff
@@ -50,27 +50,16 @@ EOF
...
-        -h)
+        h)
             usage
             exit
             ;;
-        --)
-            shift
-            break
-            ;;
         *)
```
kilroyd: We are calling getopts in normal (non-silent) mode.
That means on an unknown option, we will get '?' in $opt.
If getopts sees -o without an argument, getopts will also put '?' in $opt.

Prefer to look for '?' in this case rather than *

tests/bootstrap
```diff
@@ -50,27 +50,16 @@ EOF
...
-            echo "Unrecognized option $1"
             usage
             exit 1
             ;;
```
kilroyd: After the case, please reset the script arguments so that non-option arguments are available. We don't currently have any, but we should reset anyway.

shift $((OPTIND-1))

PR #244, opened 2019-11-26
Disable kernel_module test on macOS

a-khabarov:
Change-Id: I4ca481bf26f69b7ccab9690c5a12240733c43003


tests/kernel_module/module2/build.bp
```diff
@@ -13,4 +13,7 @@ bob_kernel_module {
     ],
     local_include_dirs: ["../module1"],
     build_by_default: true,
+    osx: {
+        enabled: false,
+    }
```
chrisdiamand: Hmm I think we might want a comma after the `}` here.

PR #246, opened 2019-11-27
Introduce new 'warning' token to the syntax.

kilroyd:
This token is used to mark a config option with a
custom warning message. The token must be followed
by a quoted string.

Change-Id: Id058b854cb7a77bc114c2228a62a24f65d107adb
Signed-off-by: Dzanan Bajgoric <dzanan.bajgoric@arm.com>

### DIFF
--- a/config_system/config_system/general.py
+++ b/config_system/config_system/general.py
@@ -185,6 +185,15 @@ def get_options_depending_on(dependent):
     return enabled_options
 
 
+def get_warning(key):
+    """
+    Returns the warning associated with the given config option.
+    Returns None if there isn't an associated warning.
+    """
+    opt = data.get_config(key)
+    return opt.get('warning', None)
+
+
 def set_initial_values():
     "Set all configuration objects to their default value"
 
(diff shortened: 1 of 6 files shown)


config_system/config_system/general.py
```diff
@@ -185,6 +185,13 @@ def get_options_depending_on(dependent):
     return enabled_options
 
 
+def get_warning(opt):
+    """Returns the warning associated with given config option or None
+    if the config options doesn't have the warning key."""
```
chrisdiamand: Should this be "if the config option" (no -s)?

config_system/config_system/general.py
```diff
@@ -185,6 +185,15 @@ def get_options_depending_on(dependent):
     return enabled_options
 
 
+def get_warning(opt):
+    """
+    Returns the warning associated with the given config option.
+    Returns None if there isn't an associated warning.
+    """
+    config = data.get_config(opt)
+    return config['warning'] if "warning" in config else None
```
chrisdiamand: Something else, sorry - can't we just use `config.get('warning')` here? (or `config.get('warning', None) if you want to be explicit)?

That's exactly what it's doing functionally so I guess we may as well use the right helper.

kilroyd: I agree, that's slightly better. I'll have the change up in a sec...

chrisdiamand: Thanks!

PR #249, opened 2019-11-27
Reduce scope of ModuleDir on Soong

chrisdiamand:
In the Soong plugin, unlike the Android.mk backend, calls to
`ModuleDir()` return the whole path from the root of the Android tree up
to the directory containing `Android.bp`. This means that setting srcdir
would be nonsensical, because the project root is included by
`ModuleDir()`.

However, this makes a few things difficult:
 - Transform source regex matching has different behaviour than on the
   other backends, because more of the path is included when matching
   the regular expression.
 - We need to know srcdir for other reasons, for example as a variable
   available to generate_source commands.

Fix the above two issues by wrapping calls to `ctx.ModuleDir()` with a
new helper, `projectModuleDir()`, which returns the location of an
`Android.bp` file relative to the project root, rather than the Android
root.

This fits quite neatly into our current handling of paths:
 - srcdir is now a path relative to the working directory.
 - After the processPaths mutator, source paths are still relative to
   the project dir.
 - Later, we convert everything into "full" paths by prefixing with
   `g.srcPrefix()` or `g.buildDir()`, as appropriate.

Change-Id: Id19163d141fc4e2b4fa9b7f87c39ba2a46ea0af2

### DIFF
--- a/bootstrap_soong.bash
+++ b/bootstrap_soong.bash
@@ -64,6 +64,7 @@ sed -e "s#@@BOB_CONFIG_OPTS@@#${BOB_CONFIG_OPTS}#" \
     -e "s#@@BUILDDIR@@#${BUILDDIR}#" \
     -e "s#@@CONFIGNAME@@#${CONFIGNAME}#" \
     -e "s#@@CONFIG_JSON@@#${CONFIG_JSON}#" \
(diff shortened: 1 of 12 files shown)


chrisdiamand:
Makes sense - looks like they're both closed now, so I'll go ahead and merge this :)

PR #260, opened 2019-12-06
Implement AndroidMk() method on kernel module and generate backends

micprz:
Change-Id: Icc6e5f9fbaa903ea3a1275ce8a3b688b17fd6635

### DIFF
--- a/core/soong_gen.go
+++ b/core/soong_gen.go
@@ -286,6 +286,20 @@ func (m *genBackend) GenerateAndroidBuildActions(ctx android.ModuleContext) {
 	m.buildInouts(ctx, args)
 }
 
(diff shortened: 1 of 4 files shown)


core/soong_gen.go
```diff
@@ -286,6 +286,20 @@ func (m *genBackend) GenerateAndroidBuildActions(ctx android.ModuleContext) {
 	m.buildInouts(ctx, args)
 }
 
+func (m *genBackend) AndroidMkEntries() android.AndroidMkEntries {
+	// we skip if multiple outputs defined, as AndroidMkEntries struct support only single one
+	// (transform_source and generate_source basically)
```
kilroyd: Can we make this clearer by replacnig the bracketed text with "This should only affect bob_transform_source and bob_generate_source"

PR #265, opened 2019-12-12
Add support for external libs on Soong

a-khabarov:
Change-Id: I6946068060ca22202533db987d5daced8286538d


tests/external_libs/external/Android.bp
```diff
@@ -0,0 +1,18 @@
+cc_library_static {
```
kilroyd: Can we name this file something other than Android.bp to avoid normal DDK builds picking this up unnecessarily. Also could cause problems if multiple Bob checkouts are present in the tree. Instead create a symlink during bootstrap for both soong and android make.

Maybe external_lib.bp ?

PR #276, opened 2020-01-09
Improve consistency of install paths tests

a-khabarov:
Testing of install paths should be consistent. This commit ensures
that each Bob module type has an instance with an install group,
updates tests/build_tests.sh to check that modules get installed
correctly, and also includes some other miscellaneous improvements
to install paths testing.

Change-Id: I9b8bbc849c9f3b35252a59050698436e49a0e131

### DIFF
--- a/tests/build.bp
+++ b/tests/build.bp
@@ -1,5 +1,5 @@
 /*
- * Copyright 2018-2019 Arm Limited.
+ * Copyright 2018-2020 Arm Limited.
  * SPDX-License-Identifier: Apache-2.0
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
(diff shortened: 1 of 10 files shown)


tests/source_encapsulation/build.bp
```diff
@@ -1,5 +1,5 @@
 /*
- * Copyright 2019 Arm Limited.
+ * Copyright 2019-2020 Arm Limited.
```
kilroyd: we haven't changed this file anymore...

tests/build.bp
```diff
@@ -17,6 +17,39 @@
...
+}
+
+bob_install_group {
+    name: "IG_host_libs",
+    builder_android_make: {
+        install_path: "$(HOST_OUT_SHARED_LIBRARIES)",
+    },
+    builder_ninja: {
+        install_path: "install/host/lib",
+    },
```
chrisdiamand: Should we have `builder_soong` values of `install_path` for these two? All the other ones have it (even though it is then ignored :) )

tests/transform_source/build.bp
```diff
@@ -91,9 +91,22 @@ bob_binary {
     srcs: ["main.cpp"],
 }
 
+bob_transform_source {
+    name: "validate_install_transform_source",
+    srcs: ["f5.in"],
+    out: {
+        match: "f5.in",
+        replace: ["validate_install_transform_source.txt"]
```
chrisdiamand: Should we put a `$1` in the `replace` value? IMO it's a bit unusual to have a constant value here - if there were multiple sources, then the fixed output file would keep getting overwritten.

PR #277, opened 2020-01-10
Fix test gen_source_depfile_with_implicit_outs (always rerun)

micprz:
Change-Id: Ie4ba998d20715363c8e2422855d15103bf709322

### DIFF
--- a/tests/generate_source/build.bp
+++ b/tests/generate_source/build.bp
@@ -218,11 +218,9 @@ bob_generate_source {
     out: ["output.txt"],
     depfile: true,
     tool: "gen_with_dep.py",
-    cmd: "${tool} -o ${out} -d ${depfile} ${in}",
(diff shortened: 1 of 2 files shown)


tests/generate_source/build.bp
```diff
@@ -218,11 +218,9 @@ bob_generate_source {
     out: ["output.txt"],
     depfile: true,
     tool: "gen_with_dep.py",
-    cmd: "${tool} -o ${out} -d ${depfile} ${in}",
+    cmd: "${tool} -o ${out} -d ${depfile} ${in} && touch ${gen_dir}/out.h",
```
kilroyd: We're testing implicit outputs (which by definition shouldn't be named on the command line).

Can we set things up so that ${tool} writes out.h based on an input flag, so that the test looks closer to the real use-case please

micprz: Changed in gerrit update.

PR #283, opened 2020-01-16
Implement bob_resource on soong

micprz:
Extracted AndroidProps and using it for resources to support optional
installing in vendor partition (instead of system partition).

Implementing our custom module to handle installing onto data partition
was necessary as soong's available prebuilt_etc doesn't support that.

Change-Id: I7c229809f1b5c102ead3843043b4e614b3f735fd

### DIFF
--- a/docs/module_types/bob_resource.md
+++ b/docs/module_types/bob_resource.md
@@ -8,6 +8,11 @@ need while executing.
 This will reference an `bob_install_group` so it gets copied to an appropriate location
 relative to the binaries.
 
+For the Soong plugin, the `install_path` set in the `bob_install_group` must be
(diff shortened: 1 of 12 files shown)


docs/module_types/bob_resource.md
```diff
@@ -50,3 +56,8 @@ Source files to copy to the installation directory.
 ----
 ### **bob_resource.add_to_alias** (optional)
 Adds this module to an alias.
+
+----
+### **bob_module.owner** (optional)
+Value to use on Android for `LOCAL_MODULE_OWNER`
+Setting this will force install onto `vendor` partition instead of `system`.
```
kilroyd: Please update this line to match the version in common_properties

PR #298, opened 2020-02-06
Add support for RSP files

chrisdiamand:
Enable lists of input files longer than the command line length by
adding support for Ninja's `rspfile` feature, which writes a file out
immediately before executing the command, which could contain the input
or output list, rather than passing it in on the command line.

Change-Id: I597406bd88831030359ab94aaa8d15d8f90cea05


chrisdiamand:
Travis is complaining that the PR merge commit doesn't have a Signed-off-by, which won't matter after a rebase - so I'll go ahead and commit this :)

PR #299, opened 2020-02-06
Only output modules for Android if required

kilroyd:
The Android builders should only include modules if they are required by the default targets.

This pull request also includes a refactoring to pull commonly used code out of the Android make backend.


kilroyd:
One of the PR tests failed because the testify repo didn't seem to respond. It was fine on the other platforms and in the branch tests, so I think this is fine to merge.

PR #300, opened 2020-02-10
Revert "Always rename Bob modules"

chrisdiamand:
This reverts commit 79a8ed4907af90ac950cf7580694cdedd6923848 - the
Android.bp backend will run Bob in a separate process, so the names will
no longer conflict with the "real" modules built by Soong.

Disable the tests on the Soong plugin; it's more useful to revert this
earlier on, rather than keeping the Soong plugin 100% working, so we
don't get extra naming complexity creeping into the Android.bp
implementation.

Other than the above modification to `tests/bootstrap_soong`, this
commit is exactly as generated by `git revert`, and then including
everything except the copyright year changes. The comment at the end of
ccModuleName() is not removed, because it still applies.

Change-Id: I315a959dcfe1133f6903c4dd9a872a182a2b9fa1


core/library.go
```diff
@@ -851,10 +851,7 @@ func (handler *graphMutatorHandler) ResolveDependencySortMutator(mctx abstr.Bott
 	if _, ok := mainModule.(*defaults); ok {
 		return // ignore bob_defaults
 	}
-
-	mainModuleName := buildbpName(mainModule.Name())
```
kilroyd: Could we keep the hoist of mainModuleName?

core/soong_gen.go
```diff
@@ -370,26 +370,21 @@ func (m *genBackend) AndroidMkEntries() []android.AndroidMkEntries {
...
+func (gc *generateCommon) getHostBinModule(mctx android.TopDownMutatorContext) (hostBin android.Module) {
 	mctx.VisitDirectDepsWithTag(hostToolBinTag, func(m android.Module) {
-		hostBinModule = m
+		hostBin = m
 	})
-	if hostBinModule == nil {
+	if hostBin == nil {
 		panic(fmt.Errorf("Could not find module specified by `host_bin: %v`", proptools.String(gc.Properties.Host_bin)))
 	}
-	bin, ok := hostBinModule.(*binary)
```
kilroyd: Useful to keep this check too?

core/library.go
chrisdiamand: Sure, I don't see why not :)

core/soong_gen.go
chrisdiamand: Yep - although I've just spotted that the panic() inside the `if` incorrectly references `bin`, which would be Nil if the check failed. So I'll leave the check, but fix that.

core/linux.go
```diff
@@ -434,10 +434,7 @@ func (l *library) GetStaticLibs(ctx blueprint.ModuleContext) []string {
 	g := getBackend(ctx)
 	libs := []string{}
 	for _, moduleName := range l.Properties.ResolvedStaticLibs {
-		dep, _ := ctx.GetDirectDep(bobName(moduleName))
-		if dep == nil {
```
kilroyd: Sorry, I didn't see this one last time around. Do we want this panic too?

chrisdiamand: Yep, reverted

PR #303, opened 2020-02-14
Make generator modules an independent Soong plugin

a-khabarov:
This commit extracts `genBackend`-related code from the main Soong plugin into
an independent plugin, making the main plugin redundant now that all of its
parts have been made into independent plugins.

There is no support for generated libraries and binaries as that would introduce
maintenance issues.

This commit also includes some refactoring to the extracted code:

* `inoutForSrc` is taken from `core/generated.go` and refactored to use
  `soongInout` instead of `inout` and `android.Path` instead of
  `sourceFilePath`.
* Tags are taken from `core/generated.go` but now disambiguated using
  Go's typesystem the way it is done in Soong.
* All references to variables from `core/soong_config.go.in` are removed.

`transform_source` test is updated to use different `export_gen_include_dirs`
on Soong, since here the sources matched by regex are prefixed not relative to
bob/tests but relative to the module directory. Since these paths will be
handled correctly by the Android.bp backend there is no reason to fix this.

Change-Id: Icc1c45fd92e4816ccf5e7a842c20b8487a3c230f

### DIFF
--- a/plugins/genrulebob/genrule.go
+++ b/plugins/genrulebob/genrule.go
@@ -1,0 +1,436 @@
+// +build soong
+
+/*
+ * Copyright 2020 Arm Limited.
+ * SPDX-License-Identifier: Apache-2.0
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package genrulebob
+
+import (
+	"fmt"
+	"path/filepath"
+	"regexp"
+	"strings"
+
+	"android/soong/android"
+	"android/soong/cc"
+	"android/soong/genrule"
+
+	"github.com/ARM-software/bob-build/internal/utils"
+
+	"github.com/google/blueprint"
+)
+
+type MultiOutProps struct {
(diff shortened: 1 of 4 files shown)


plugins/genrulebob/genrule.go
```diff
@@ -0,0 +1,413 @@
...
+		for _, prop := range props {
+			new_props = append(new_props, re.ReplaceAllString(source.Rel(), prop))
+		}
+		return
+	}
+	mop := m.Properties.Multi_out_props
+
+	sio.in = android.Paths{source}
+	sio.out = pathsForModuleGen(ctx, replaceSource(mop.Replace))
+	sio.implicitSrcs = android.PathsForModuleSrc(ctx, replaceSource(mop.Implicit_srcs))
```
chrisdiamand: I'd expect implicit sources derived from a specific source file to be relative to the same directory as `source`.
Mostly this will be the module's source dir - so `PathsForModuleSrc` would be correct. However, if source has come from another generated module, this will be somewhere in the build dir instead.

Can we rewrite this as something along the lines of `filepath.Join(source.NonRel(), replaceSource(mop.Implicit_srcs))`? Note that you might have to do something weird to get the non-rel component of `source`, as `paths.go` doesn't provide a built-in way to do this.

chrisdiamand: I'm not sure `PathsForSource` will work either - it's closer, but won't allow us to use paths from the `out` directly:
```
// PathForSource joins the provided path components and validates that the result
// neither escapes the source dir nor is in the out dir.
// On error, it will return a usable, but invalid SourcePath, and report a ModuleError.
func PathForSource(ctx PathContext, pathComponents ...string) SourcePath {
```
See "nor is in the out dir"...

Maybe you could check explicitly for if `source` is an `android.OutputPath`, and when it is, use something like `OutputPath.WithoutRel().Join(...)`?

chrisdiamand: Thanks!

PR #317, opened 2020-02-20
Path helpers

kilroyd:
Add trivial path helpers which make callers slightly easier to read.


chrisdiamand:
These are pretty straightforward so let's get this PR closed :)

PR #320, opened 2020-02-26
Path helpers II

kilroyd:
Fixup the path helper for the script directory so that it can use build system path variables.
Rename sourcePrefix() for consistency.
Add documentation for the Android backend path helpers.


kilroyd:
The failure in the pull request build is a failure in the sign-off checker. This should be OK to merge.

PR #327, opened 2020-02-27
Deprecate export_[ldlibs|shared_libs|static_libs]

a-khabarov:
`export_*libs` properties negatively affect memory consumption, are
redundant and decrease clarity. They will be completely removed in the
near future.

This commit merges these properties with their non-export variants.
The changes do not break existing `.bp` files, however, they should be
updated asap to use the non-export variants of these properties.

Change-Id: I726c641dc67baf8d815e376de00c8c0adc0e3ad0

### DIFF
--- a/core/library.go
+++ b/core/library.go
@@ -56,24 +56,26 @@ type BuildProps struct {
 	Library_version string
 
 	// The list of shared lib modules that this library depends on.
-	// 'shared_libs' only makes sense for share libraries and executables -
-	// static libs should use 'export_shared_libs', so the dependency is
-	// propagated to the top-level object.
+	// These are propagated to the closest linking object when specified on static libraries.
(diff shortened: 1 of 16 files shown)


core/library.go
```diff
@@ -56,24 +56,26 @@ type BuildProps struct {
 	Library_version string
 
 	// The list of shared lib modules that this library depends on.
-	// 'shared_libs' only makes sense for share libraries and executables -
-	// static libs should use 'export_shared_libs', so the dependency is
-	// propagated to the top-level object.
+	// These are propagated to the closest linking object when specified on static libraries.
```
kilroyd: Please update docs/module_types/common_module_properties.md for shared_libs, static_libs and ldlibs so that it contains the same text. We can skip updating the documentation on the export properties, as we will delete them shortly

PR #330, opened 2020-03-02
Refactor handling of Android.bp.in for plugins

micprz:
Instead of creating separate plugins/Android.bp from bootstrap script,
we combine it with main Android.bp containing all modules. That way we
have only single file.

Use project-unique identifier in definition of bootstrap go packages,
enabling us to have multiple projects using plugins in parallel, without
overwriting then-shared plugins/Android.bp.
Use hash of source dir path to generate such id, so we don't need to
depend on other env variable such as PROJ_NAME.

Change-Id: Ie3d43034f839fb4fe8102a7f10b82eab7e23b7de

### DIFF
--- a/core/androidbp_backend.go
+++ b/core/androidbp_backend.go
@@ -18,6 +18,9 @@
 package core
 
 import (
+	"crypto/sha1"
+	"encoding/hex"
+	"io/ioutil"
 	"path/filepath"
 	"strings"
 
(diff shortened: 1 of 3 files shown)


core/androidbp_backend.go
```diff
@@ -89,10 +92,33 @@ func androidBpSingletonFactory() blueprint.Singleton {
...
+	// read definitions of plugin packages
+	content, err := ioutil.ReadFile(filepath.Join(getBobDir(), "plugins/Android.bp.in"))
+	if err != nil {
+		utils.Exit(1, err.Error())
+	}
+
+	// use source dir to get project-unique identifier,
+	// generate 10 chars in hex based on its hash
+	h := sha1.New()
+	h.Write([]byte(getSourceDir()))
```
chrisdiamand: OK - this works, because `srcdir` returns the path relative to the working directory, which on `Android.bp` is the root of  the Android tree. So e.g. we'd be hashing something like `external/bob-build/tests` here, which is enough to identify the project :)

PR #343, opened 2020-03-18
Implement utils.Expand function

a-khabarov:
Change-Id: I8abb90627c326c916a37011edaaf3051a84a5d09

### DIFF
--- a/internal/utils/utils.go
+++ b/internal/utils/utils.go
@@ -288,3 +288,49 @@ func Exit(exitCode int, err string) {
 func FlattenPath(s string) string {
 	return strings.Replace(s, "/", "__", -1)
 }
+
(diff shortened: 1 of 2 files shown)


internal/utils/utils.go
```diff
@@ -288,3 +288,53 @@ func Exit(exitCode int, err string) {
 func FlattenPath(s string) string {
 	return strings.Replace(s, "/", "__", -1)
 }
+
+// Expand function is similar to os.Expand, with two differences:
+// * "$$" is always converted to "$"
```
kilroyd: If we're leaving things unchanged if we don't recognize the expansion, then shouldn't we drop this $$ behaviour.
We only need it if we always replaced $anything, to stop the expansion.

PR #354, opened 2020-03-26
Add branch for blueprint to gitmodules

kilroyd:
Change-Id: Ifa10df81b323119ac823d3a021a7e0eb346b5803


kilroyd:
The macOS pull request build failed, looks like this is due to infrastructure. I don't expect this commit can have caused this, so merging.

PR #356, opened 2020-04-07
Support `out` in generated libraries

chrisdiamand:
Allow generated libraries to use a different output filename than the
module name. This allows e.g. a library which is generated via a
sub-build on Linux to be linked via a `bob_external_library` with the
same `.so` name on Android.

Change-Id: Iadef232e975ed0ed23e0b6dabc97d545931f502b

### DIFF
--- a/tests/generate_libs/build.bp
+++ b/tests/generate_libs/build.bp
@@ -130,11 +130,60 @@ bob_generate_binary {
     },
 }
 
+bob_generate_shared_library {
+    name: "libblah_shared_rename",
+    out: "libblah_shared2",
+    srcs: ["libblah/libblah.c"],
+    headers: [
+        "include/libblah.h",
+        "include/libblah_feature.h",
+    ],
+    export_gen_include_dirs: ["include"],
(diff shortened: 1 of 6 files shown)


tests/generate_libs/build.bp
```diff
@@ -130,11 +130,56 @@ bob_generate_binary {
...
+        enabled: false,
+    },
+}
+
+bob_generate_source {
+    name: "check_renamed_gen_library",
+    out: ["success.txt"],
+    module_srcs: ["binary_linked_to_gen_shared_rename"],
+    /* Check that the binary has a NEEDED entry for the renamed shared library */
+    cmd: "{ objdump -p ${in} | grep NEEDED | grep -q libblah_shared2.so ; } && touch ${out}",
```
kilroyd, line 169: Can we do this check in a python script to minimize the dependence on shell character handling?

Do we even need to do this check here? Why not just add a check to build_tests.sh to check for the presence of libblah_shared2.so?

chrisdiamand: Sure. Although I think really we should be able to guarantee passing at least a small amount of syntax correctly to the shell, otherwise we end up with scripts for even the most trivial things.

The use case for this requires that the .so name and DT_NEEDED entries match the value of `out` - so I thought it would make the most sense to add it to the test.

chrisdiamand: Done! Had to add another method (other than `objdump`) to get the tests working on OS X too.

PR #359, opened 2020-04-15
Allow setting LD in kernel module builds

chrisdiamand:
Add a new property, `kernel_ld`, which maps to the `LD` Make variable
when build kernel modules. When this is not specified, but the kernel
has CONFIG_LD_IS_LLD enabled, default this to `ld.lld`. This allows
modules compiled against kernels with LTO enabled to build.

Change-Id: I233884e8fcdfcd8eacdc485a6fce687295190c43

### DIFF
--- a/scripts/kmod_build.py
+++ b/scripts/kmod_build.py
@@ -68,7 +68,8 @@ def build_module(output_dir, module_ko, kdir, module_dir, make_command, make_arg
     # Sanitize the environment - we should only use build options passed in via
     # the command line.
     env = dict(os.environ)
-    for var in ["ARCH", "CROSS_COMPILE", "CC", "HOSTCC", "CLANG_TRIPLE", "KBUILD_EXTRA_SYMBOLS"]:
+    for var in ["ARCH", "CROSS_COMPILE", "CC", "HOSTCC",
+                "CLANG_TRIPLE", "KBUILD_EXTRA_SYMBOLS", "LD"]:
         env.pop(var, None)
 
     try:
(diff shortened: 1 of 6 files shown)


scripts/kmod_build.py
```diff
@@ -68,7 +68,7 @@ def build_module(output_dir, module_ko, kdir, module_dir, make_command, make_arg
     # Sanitize the environment - we should only use build options passed in via
     # the command line.
     env = dict(os.environ)
-    for var in ["ARCH", "CROSS_COMPILE", "CC", "HOSTCC", "CLANG_TRIPLE", "KBUILD_EXTRA_SYMBOLS"]:
+    for var in ["ARCH", "CROSS_COMPILE", "CC", "HOSTCC", "CLANG_TRIPLE", "KBUILD_EXTRA_SYMBOLS", "LD"]:
```
kilroyd: The kernel make doesn't read LD from an environment variable (it's assigned with '='), so we can drop this. Actually, this is the same for CC and HOSTCC

chrisdiamand: In the main `Makefile` I think you're correct. But there's stuff like this in various other sub-make files and helper scripts:

- `tools/lib/subcmd/Makefile:12` has `CC ?= $(CROSS_COMPILE)gcc`
- `tools/perf/util/setup.py:8` has `cc = getenv("CC")
- `tools/build/Makefile` and `tools/objtool/Makefile` both have `HOSTCC ?= gcc`

I haven't looked into exactly how these are invoked, but wouldn't it be safer just to remove potential conflicting values from the environment just in case?

kilroyd: Hmm. That's annoying. Let's commit as is then.

PR #361, opened 2020-04-16
config_system: report config inconsistencies after plugin

kilroyd:
Plugins may set config values which command line options may depend
on. So we should only complain that the output config is consistent
after the plugins have executed.

eg.

config YELLOW
    bool "Yellow"
    depends on COLOR_TV

config COLOR_TV
    bool "Color TV"
    default n

$ ./update_config.py YELLOW=y
ERROR: YELLOW=y was ignored; its dependencies were not met: COLOR_TV[=n]

... will fail if COLOR_TV is set by a plugin

We want the configs to be as consistent as possible for the plugins.
So continue to run enforce_dependent_values before, however disable
any errors in this run. Rerun after the plugins execute to recheck
consistency, and enable error reporting in this final run.

Note this change converts an exception to an error in
enforce_dependent_values so that we can report multiple issues in each
run.

Add a test to check this behavior. In doing so, ensure that the tests
are resetting the error counts on each call to update_config.py

Change-Id: Ic1ecac00c393bb61a28dc1245a6012b7994d7877

### DIFF
--- a/config_system/config_system/general.py
+++ b/config_system/config_system/general.py
@@ -39,30 +39,55 @@ def is_visible(config):
     return expr.condexpr_value(e)
 
 
-def enforce_dependent_values(auto_fix=False):
+def enforce_dependent_values(stage, fix_bools=False,
+                             error_level=logging.WARNING):
     """
     Check that values are consistently set, specifically with respect
     to whether the dependencies are correct.
 
-    For boolean values we only reset to n if auto_fix is
-    True. Otherwise we raise an exception (the Mconfig needs to be
-    fixed).
+    This function is called when we expect the values to be
+    consistent. i.e. after reading in a new config, or prior to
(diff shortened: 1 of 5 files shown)


config_system/config_system/general.py
```diff
@@ -39,30 +39,55 @@ def is_visible(config):
...
 
-    For boolean values we only reset to n if auto_fix is
-    True. Otherwise we raise an exception (the Mconfig needs to be
-    fixed).
-
-    For string and int values, if the configuration value is not
-    enabled, reset them to "" and 0 respectively. This is different to
-    the bool case as we need to reset default values.
+    This function is expected to be called when we expect the values
+    should be consistent. i.e. after reading in a new config, or prior
```
dliviu, line 49: This reads weird. Maybe something like "This function should be called when we expect the values to be consitent"?

PR #362, opened 2020-04-23
Remove TOPNAME option from Bob

a-khabarov:
This commit removes the `TOPNAME` option from Bob.
Blueprint's `TOPNAME` is now always being set to "build.bp".

Change-Id: Ia6a24f32d5b623591dcc7edf927b4f34035892ba

### DIFF
--- a/bootstrap.bash
+++ b/bootstrap.bash
@@ -1,6 +1,6 @@
 #!/bin/bash
 
-# Copyright 2018-2019 Arm Limited.
+# Copyright 2018-2020 Arm Limited.
 # SPDX-License-Identifier: Apache-2.0
 #
 # Licensed under the Apache License, Version 2.0 (the "License");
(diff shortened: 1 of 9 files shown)


bootstrap.bash
```diff
@@ -89,6 +83,7 @@ BOB_DIR="$(relative_path $(pwd) "${SCRIPT_DIR}")"
 CONFIG_FILE="${BUILDDIR}/${CONFIGNAME}"
 CONFIG_JSON="${BUILDDIR}/config.json"
 
+export TOPNAME="build.bp"
```
kilroyd, line 86: Please update the BOOTSTRAP_VERSION in this commit, as this causes an incompatibility.

a-khabarov:
Travis is giving an error for this commit: https://travis-ci.com/github/ARM-software/bob-build/jobs/322494384. Looks like it takes bootstrap.bash from the top of the master branch (the one that errors when TOPNAME is not set) while taking tests/bootstrap_*  from the commit itself, causing the build to error with "TOPNAME must be set". The test needs fixing.

PR #370, opened 2020-05-07
Add support for testing the availability of compiler flags

dliviu:
Newer versions of a compiler adds support for new flags that would not be
understood by older compilers. To allow for the addition of those new flags
into the build.bp file, add '{{addIfSupported' late template
that will take as an argument the compiler flag to be tested for support
by the current toolchain compiler. Note that 'cflags' and 'export_cflags'
are tested with both the C and C++ compilers.

Signed-off-by: Liviu Dudau <liviu.dudau@arm.com>
Change-Id: Ifce31bb3d7df44d37ba0b16a72e8475980e5691e

### DIFF
--- a/tests/flag_supported/test.c
+++ b/tests/flag_supported/test.c
@@ -1,0 +1,15 @@
+#include <stdio.h>
+#include <stdlib.h>
+
+void useless(void)
+{
+    printf("I am not used\n");
+}
+
+const char main(int argc, char** argv)
+{
+    (void)argc;
+    (void)argv;
+    const int i = 0;
+    return i;
+}
--- a/core/late_template.go
+++ b/core/late_template.go
@@ -129,3 +129,49 @@ func matchSourcesMutator(mctx blueprint.TopDownMutatorContext) {
 		}
 	}
 }
+
+var flagFilterRegex = regexp.MustCompile(`\{\{add_if_supported\s+(.+?)\}\}`)
+
+// Iterate over an array of flags and filter out the ones that are not supported
+// based on the list of languages in a toolchain
+func checkFlags(flags, languages []string, tc toolchain) []string {
+	out := []string{}
+	for _, flag := range flags {
+		matches := flagFilterRegex.FindAllStringSubmatch(flag, -1)
+		if matches != nil {
+			for _, match := range matches {
+				for _, lang := range languages {
+					if tc.checkFlagIsSupported(lang, match[1]) {
+						out = append(out, match[1])
+						break
+					}
+				}
+			}
+		} else {
+			out = append(out, flag)
+		}
+	}
+	return out
+}
+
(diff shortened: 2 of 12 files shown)


tests/flag_supported/test.c
```diff
@@ -0,0 +1,15 @@
+#include <stdio.h>
+#include <stdlib.h>
+
+void useless(void)
+{
+    printf("I am not used\n");
+}
+
+const char main(int argc, char** argv)
```
chrisdiamand, line 9: That's a funky looking `main` signature :)

Travis is complaining that the return value isn't `int` - but also isn't it more common to write `char **argv`? The `*` only applies to one argument, so that style avoids misleading declarations like `char** a, b`, where `a` and `b` look like they should both be `char **`, but `b` is just a `char`.

dliviu: The idea is to trigger a lot of warnings from the normal run of a compiler without the additional flags so that I can check that the right flags get added in order to get a clean build.
The funky main signature was intentional as a quick and easy way of generating some warnings in the newer compilers that started to care about how your `main` looks like.

chrisdiamand: D'oh, sorry, I completely missed the point of the test.

I guess there are two issues with the travis runs then:
```
FAILED: build-in-src/target/objects/bob_test_flag_supported/flag_supported/test.c.o 

 gcc -c -Wall -Werror -Wno-discarded-qualifiers -Wno-ignored-qualifiers -Wno-main-return-type -MMD -MF build-in-src/target/objects/bob_test_flag_supported/flag_supported/test.c.o.d flag_supported/test.c -o build-in-src/target/objects/bob_test_flag_supported/flag_supported/test.c.o

flag_supported/test.c:9:12: error: return type of ‘main’ is not ‘int’ [-Werror=main]

 const char main(int argc, char** argv)

            ^

cc1: error: unrecognized command line option ‘-Wno-main-return-type’ [-Werror]

cc1: all warnings being treated as errors
```
- The return type error is caused by `-Werror=main` on this version, so do we need to try and add `add_if_supported -Wno-error=main` too?
- `Wno-main-return-type` is being added to the flags, even though it doesn't appear to be supported, which would suggest that there's a bug in the `add_if_supported` checking somewhere :(

dliviu: Found the two issues: old gcc uses '-Wmain' for warning about 'main' signature, so I've added '-Wno-main' to the flags. As for the invalid flags being added to the supported flags, it is due to the special way 'gcc' handles the '-Wno-<flag>' as command line parameter. If there is no actual file to compile 'gcc' will not warn about the parameter being unknown, but it will as soon as you pass a real file to compile. I've fixed it by stripping the 'no-' part from the flag and testing that with the compiler to see if it is supported.

PR #371, opened 2020-05-07
Add LC_ALL=C to bplist_regen.sh to ensure consistent ordering.

dliviu:
Without a fixed locale the output of sort will depend on the locale
settings on the machine where 'bplist_regen.sh' gets run, which
might re-order some of the entries in the old list.

Signed-off-by: Liviu Dudau <liviu.dudau@arm.com>
Change-Id: I1dd4c80f87a4a72d9e3a8b8702e314b929a2aafa

### DIFF
--- a/tests/bplist_regen.sh
+++ b/tests/bplist_regen.sh
@@ -1,6 +1,6 @@
 #!/bin/sh
 
-# Copyright 2018 Arm Limited.
+# Copyright 2018, 2020 Arm Limited.
 # SPDX-License-Identifier: Apache-2.0
 #
 # Licensed under the Apache License, Version 2.0 (the "License");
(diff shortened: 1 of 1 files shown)


tests/bplist_regen.sh
```diff
@@ -36,6 +36,7 @@ find . -mindepth 1 \
 echo ./bob/Blueprints >> "${TEMP_LIST_FILE}"
 echo ./bob/blueprint/Blueprints >> "${TEMP_LIST_FILE}"
 
+LC_ALL=C
```
chrisdiamand, line 39: Does this need to be exported?

On my machine, I don't have `LC_ALL` set at all (i.e. it's not already exported but set to another value).

This means if I run `bplist_regen.sh`, even with this change, it gets ordered strangely, with `Blueprints` after `blueprint`.

Changing to `export LC_ALL=C` fixes this.

tests/bplist_regen.sh
```diff
@@ -1,6 +1,6 @@
 #!/bin/sh
 
-# Copyright 2018 Arm Limited.
+# Copyright 2018,2020 Arm Limited.
```
chrisdiamand, line 3: Being really nitpicky... Any chance you could put a space before `2020` please? :) thanks!

tests/bplist_regen.sh
kilroyd: Alternatively, just prefix the sort command with this (it should be the only relevant commant that needs it)

dliviu: I've updated the patch to prefix the command. I don't think the script should export this.

tests/bplist_regen.sh
dliviu: Done.

PR #374, opened 2020-05-11
Add proper support for property exclude_implicit_srcs

micprz:
Add documentation and test. Test now uses external join_srcs.py script
to actually read input files and produce relevant output. Adapted
previous test to use it too.

Change-Id: Iaaa18b48e900e116ea25c1aa50e9925ba4fa9fee

### DIFF
--- a/tests/generate_source/bn.src
+++ b/tests/generate_source/bn.src
@@ -1,0 +1,1 @@
+#error This file should not be included in compilation
--- a/tests/generate_source/build.bp
+++ b/tests/generate_source/build.bp
@@ -238,7 +238,18 @@ bob_generate_source {
     name: "gen_source_globbed_implicit_sources",
     implicit_srcs: ["*.implicit.src"],
     out: ["validate_globbed_implicit_dependency.c"],
-    cmd: "echo 'int main(void) { return 0; }' > ${out}",
+    tool: "join_srcs.py",
+    cmd: "python ${tool} --src-dir ${module_dir} --use-a --out ${out}",
+    build_by_default: true,
+}
+
+bob_generate_source {
+    name: "gen_source_globbed_exclude_implicit_sources",
+    implicit_srcs: ["*.src"],
+    exclude_implicit_srcs: ["an.implicit.src", "bn.src"],
+    out: ["validate_globbed_exclude_implicit_dependency.c"],
+    tool: "join_srcs.py",
+    cmd: "python ${tool} --src-dir ${module_dir} --use-c --out ${out}",
     build_by_default: true,
 }
 
@@ -246,6 +257,7 @@ bob_binary {
     name: "use_miscellaneous_generated_source_tests",
     generated_sources: [
         "gen_source_globbed_implicit_sources",
+        "gen_source_globbed_exclude_implicit_sources",
     ],
 }
 
--- a/tests/generate_source/join_srcs.py
+++ b/tests/generate_source/join_srcs.py
@@ -1,0 +1,46 @@
+#!/usr/bin/env python
+
+# Copyright 2020 Arm Limited.
+# SPDX-License-Identifier: Apache-2.0
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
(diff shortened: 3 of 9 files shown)


tests/generate_source/bn.src
```diff
@@ -0,0 +1 @@
+#error This file should not be included in compilation
```
chrisdiamand, line 1: Can you add end-of-file newlines to `an.implicit.src`, `bn.src` and `cn.src` please? (Github's showing that these files do not have them)

tests/generate_source/build.bp
```diff
@@ -238,14 +238,26 @@ bob_generate_source {
...
-    cmd: "echo 'int main(void) { return 0; }' > ${out}",
+    tool: "join_srcs.py",
+    cmd: "python ${tool} --src-dir ${module_dir} --use-a --out ${out}",
+    build_by_default: true,
+}
+
+bob_generate_source {
+    name: "gen_source_globbed_exclude_implicit_sources",
+    implicit_srcs: ["*.src"],
+    exclude_implicit_srcs: ["an.implicit.src", "bn.src"],
```
chrisdiamand, line 249: How does this test verify that these implicit sources have been excluded?

tests/generate_source/join_srcs.py
```diff
@@ -0,0 +1,42 @@
+#!/usr/bin/env python
+
+# Copyright 2018-2020 Arm Limited.
```
chrisdiamand, line 3: 2018?

tests/generate_source/join_srcs.py
```diff
@@ -0,0 +1,42 @@
...
+    inputs.append(os.path.join(args.src_dir, 'an.implicit.src'))
+if args.use_c:
+    inputs.append(os.path.join(args.src_dir, 'cn.src'))
+
+with open(args.output, 'w') as outfile:
+    for infile in inputs:
+        if not os.path.exists(infile):
+            print("Input file doesn't exist: " + infile)
+            sys.exit(1)
+        outfile.write(open(infile, 'r').read() + '\n')
```
chrisdiamand, line 42: Would it be more "pythonic" to `try/catch` the error?

```
try:
    with open(infile, 'r').read():
        ...
except (IOError, OSError) as e: # IIRC reading from a file can throw either of these
    print(e)
```

And it looks like you're adding the newlines here? Is there a particular reason for that? It seems like it's just compensating for the fact that your text editor hasn't added them to the source files, unless I've missed something and it's part of the test somehow?

tests/generate_source/build.bp
micprz: In two ways:
1. If an.implicit.src will be pulled in, it will fail because resulting binary already contain main() function.
2. If bn.src will be pulled in, #error directive will cause compile to fail.

tests/generate_source/join_srcs.py
micprz: I add newlines to not depend on input files having them or not.

micprz: Because it is possible to join 2 input files with this script.

tests/generate_source/join_srcs.py
micprz: Doh...will fix it

tests/generate_source/bn.src
micprz: ok

tests/generate_source/build.bp
chrisdiamand: OK, that tests that `join_srcs.py` doesn't accidentally use more than one implicit source.

But it doesn't test the `exclude_implicit_srcs` property, because `implicit_srcs` never gets written to the command line anyway! So even if the excludes were completely ignored, this would still pass :(

In order to check that these files are actually being excluded, we'd need to either manually check the `build.ninja`, or potentially do something in `build_tests.sh`: e.g. `touch an.implicit.src bn.src`, then rerun the build and check that `gen_source_globbed_exclude_implicit_sources` hasn't been rebuilt.

We can do that in a follow-up commit though so I'll approve this review for now :)

PR #376, opened 2020-05-13
Test external shared libs with zlib

chrisdiamand:
The existing external_lib tests are difficult to run outside of an
Android build, because it would be necessary to invoke another build
system somehow to build the test libraries.

To test this feature on Linux, add a simple test using the
commonly-available zlib library instead. This also requires adding the
missing options for pkg-config support to Mconfig.

`external_libs/build.bp` is also missing a copyright header - add one.

Change-Id: Id0cb02c6e3a6e9e546a66c2b9ec1758aa9198053

### DIFF
--- a/tests/Mconfig
+++ b/tests/Mconfig
@@ -159,8 +159,20 @@ source "bob/mconfig/toolchain.Mconfig"
 
 endmenu
 
+## pkg-config configuration sub-menu
+
+menu "pkg-config configuration"
+	depends on ALLOW_HOST_EXPLORE
+	help
+	  pkg-config is a Linux tool that allows discovery of available
+	  libraries. On a standard Linux install pkg-config will be
+	  setup correctly for the build host. To use pkg-config for a
+	  target system in a cross compile, the package information
+	  for the target needs to be available.
+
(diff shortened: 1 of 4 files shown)


chrisdiamand:
Travis is failing in the bootstrap version test:
```
FAILED: bootstrap_test/target/executable/use_external_zlib 

 g++ bootstrap_test/target/objects/use_external_zlib/external_libs/zlib.c.o -o bootstrap_test/target/executable/use_external_zlib -Wl,--as-needed  -Lbootstrap_test/target/shared -Wl,-rpath-link,bootstrap_test/target/shared 

bootstrap_test/target/objects/use_external_zlib/external_libs/zlib.c.o: In function `main':

zlib.c:(.text+0x42): undefined reference to `deflateInit_'

collect2: error: ld returned 1 exit status

ninja: build stopped: subcommand failed.

<------------- run_bootstrap_test.sh failed
```
This is because `pkg-config` isn't rerun between `buildme` invocations. That's kind of expected, and this is only adding a test, so incrementing the bootstrap version doesn't make sense. So I think we should ignore that failure for this change.

PR #405, opened 2020-07-09
Regenerate Android.bp on plugin build template change

kilroyd:
When Android.bp.in changes, the Android.bp file needs to be
regenerated

Change-Id: Icba33adfb3299156b3087dd3d65375b410c90d55

### DIFF
--- a/core/androidbp_backend.go
+++ b/core/androidbp_backend.go
@@ -210,7 +210,9 @@ func (s *androidBpSingleton) GenerateBuildActions(ctx blueprint.SingletonContext
 	sb := &strings.Builder{}
 
 	// read definitions of plugin packages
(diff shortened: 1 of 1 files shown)


kilroyd:
> Hmmm... isn't Android.bp.in converted to Android.bp during `bootstrap_androidbp`? That's earlier than generating the ninja file, so how does adding this dependency help?

We no longer copy Android.bp.in to Android.bp during Bob bootstrap. That was changed to output Android.bp.in as part of the main Android.bp being writted by Bob. This happens shortly after the modified code (after inserting a hash and some text replacement).

chrisdiamand:
Ah yeah I see, in `androidBpSingleton.GenerateBuildActions`.

PR #426, opened 2020-09-04
Helpful error message for overridden kernel config options.

birunts:
There is possibility that Bob will try to override options already
set in the kernel's config using EXTRA_CFLAGS. This causes
'macro redefined' error with curt message.

Check for possible option overriding and print more descriptive
error message.

Change-Id: Ic908ffc85b2b6d1c39e461e8673e04cbeaf717f2


chrisdiamand (approved):
Awesome thanks!

PR #427, opened 2020-09-04
Clear source on choice group configs.

birunts:
While resetting configuration by menuconfig 'source' property
is not properly cleared for configs in a choice group.

Change-Id: I580ed60041375a99984aa66ac48050e1a8af5138


kilroyd (approved):
Thanks!

PR #428, opened 2020-09-15
Improved encapsulation tests

chrisdiamand:
Add new encapsulation tests which check the contents of `module_srcs`
and `${modname_out}`, as well as incorrect inclusion of include
directories generated by transitive *non*-encapsulated dependencies.

*** These currently fail ***

The `${modname_out}` property should include ${modnames}'s encapsulated
dependencies' outputs, but currently does not. This will be fixed in a
subsequent commit.

Change-Id: I5eb02ce0f7450651fe1f282020f3a9a5a4aff2dc


birunts (approved):
Thanks!

PR #429, opened 2020-09-29
Fixed encapsulation output propagation.

birunts:
The `${modname_out}` property and contents of the `module_srcs` should contain
${modnames}'s encapsulated dependencies' outputs as well as correctly include
directories generated by transitive *non*-encapsulated dependencies.

Correct propagated outputs based on `encapsulates` property.

Enable new encapsulation tests which verifies correctness of encapsulation.
Fix previously written tests which have tested wrongly defined encapsulation
feature.

The outputs for the command are set in 'blueprint.BuildParams: Outputs'
for the module. We use 'm.Properties.Out' which comes from provided 'inouts.outs'.
Thus cmdOutputs() was only being used for dependency tracking, which we
don't need to do. Remove cmdOutputs() method for encapsulatedOutputProducer.

Change-Id: I4c4c2e187ec77f91be54d40ca699562f52752c83


kilroyd (approved):
I think we can ignore the python3-pip install errors. If they recur we may need to look at updating our test configurations

PR #434, opened 2020-10-09
Remove SplittableProps from target-specific props

chrisdiamand:
Now that modules can use multiple combined types as their features and
top-level properties, remove `host_supported` and `target_supported`
from the structs used inside `host: {}` and `target: {}` blocks.

As a result, we no longer need to check that they aren't accidentally
set, because the properties simply do not exist.

Change-Id: I22e383a6c0f18820e8e037e61d4ffab8abf0ee12

### DIFF
--- a/core/defaults.go
+++ b/core/defaults.go
@@ -59,7 +59,7 @@ func (m *defaults) build() *Build {
 }
 
 func (m *defaults) topLevelProperties() []interface{} {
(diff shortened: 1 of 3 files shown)


birunts (approved):
It's OK. Thanks!

kilroyd (approved):

PR #435, opened 2020-10-09
Type-independent target-specific properties

birunts:
Target specific data (for host and target) is hardcoded to BuildProps
structure which makes module always to have the same property sets.

Change target type to be type-independent which uses reflection
to put the right type inside.

Change-Id: If3bd60099ffb1a9a27520043b283b199952f0b56


kilroyd (approved):
LGTM, thanks!

PR #442, opened 2020-10-15
Declare Bob scripts using StaticVariables

kilroyd:
The script location is then only defined once in the build.ninja. Use
CommandDeps to add the implicit dependency on the script.

For the kernel modules, kmod_build is returned in args from
generateKbuildArgs so we need to delete it from the map.

Change-Id: Icfc35d9aaad086526fc0da1afca769ac6985fdfb


birunts (approved):
It's OK, thanks.

PR #444, opened 2020-10-20
Rename bob configuration file.

birunts:
Currently bob configuration file `config.json` can be wrongly decoded
and make a collision with main configuration file `bob.config`.
Such file is only used internally by bob and should not affect
anything else that wants a configuration file in JSON format.
Rename file form `config.json' to `.bob-config.json` thus only
bob should deal with it.

Additionally update bootstrap number that will impose re-bootstrap.

Change-Id: I6af5746d2d62ef1f718f2e0985c5f0666dc80fd2

### DIFF
--- a/bob.bootstrap.version
+++ b/bob.bootstrap.version
@@ -1,2 +1,2 @@
 # Version number used to identify whether the user needs to re-boostrap their build directory.
-BOB_VERSION="6"
+BOB_VERSION="7"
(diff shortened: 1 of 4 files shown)


chrisdiamand (approved):
LGTM, thanks!

kilroyd (approved):
Thanks!

PR #449, opened 2020-10-23
Rework applying default properties to more generic way.

birunts:
Currently every module which references `bob_defaults` has to be
equipped with the same set of properties no matter if it uses them
or not. This forces e.g. all library modules to contain kernel
properties.

Change the way of applying default properties to be more generic
and which allows multiple modules to have different set of
defaultable properties.

Remove unrelated `features()` method of `defaultable` interface.

Make `generateCommon` to be `defaultable` as it uses other defaults
by `flag_defaults` property.

Change-Id: Iab922fab7eb72d0cf676aa6103e26d2e8a6a30e3

### DIFF
--- a/core/defaults.go
+++ b/core/defaults.go
@@ -58,6 +58,10 @@ func (m *defaults) build() *Build {
 	return &m.Properties.Build
 }
 
+func (m *defaults) defaultableProperties() []interface{} {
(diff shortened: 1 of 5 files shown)


kilroyd (approved):
Cool!

PR #452, opened 2020-10-27
Check error while globbing

birunts:
Currently when globbing the returned error is simply omitted.
If in parsed directory will be dangling symlink this produces no
error but creates violated build.ninja file.

There is a need to check the error and exit if it appears.

Check the error while globbing and report error of the module.

Change-Id: I08e813ca23851f8dc07c731e0e12cbedd84371f6


kilroyd (commented):
Really minor - would you mind taking the period (.) off the commit title? We don't tend to have that.

Otherwise this change looks good

kilroyd (approved):

PR #453, opened 2020-10-28
Depend on shared libs via their toc file

kilroyd:
Depend on toc files to avoid rebuilding callers when the API is
unchanged. When we do this we just need to change the implicit
dependencies on the shared libraries to an implicit dependency on the
toc file.
    
Allow disabling this functionality by setting the environment variable
BOB_ALWAYS_LINK_SHARED_LIBS=1. This allows people who encounter
problems with this mode of operation to work around it while the
majority of developers get the benefits automatically.


birunts (approved):
LGTM! Thanks

dliviu (approved):

PR #457, opened 2020-11-03
Rework kernel module properties

birunts:
Kernel properties are currently closely tied to BuildProps
which forces most of the modules to contain kernel properties
even when they don't need it at all.

Move kernel properties out of BuildProps and create new
KernelProps struct for them. The modules which need such
properties should only incorporate it.

Change-Id: I89f27d7dc466e790247806657beba382601379b4


kilroyd (approved):
Nice

PR #460, opened 2020-11-05
Fix version script paths on Soong

a-khabarov:
This commit updates `library.go` to avoid using
`getBackendPathInSourceDir` when computing version script paths
on the Android.bp backend as Soong adds the prefix automatically.

Change-Id: Ifa7428bef6056fd4c25fa86b49cc108560007f16

### DIFF
--- a/core/library.go
+++ b/core/library.go
@@ -20,6 +20,7 @@ package core
 import (
 	"errors"
 	"fmt"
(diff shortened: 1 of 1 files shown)


chrisdiamand (approved):
Seems OK to me!

kilroyd (approved):

PR #463, opened 2020-11-13
Save config directory in .bob.bootstrap

francois-berder:
This commit implements the feature to save configuration files in its own independent directory, which can be in the source tree.
By default, the configuration directory points to the build directory. Hence, we cannot assume anymore that these configuration files are located in the build directory.

### DIFF
--- a/example/bootstrap_androidbp.bash
+++ b/example/bootstrap_androidbp.bash
@@ -40,6 +40,7 @@ Usage:
   profiles and explicit options (like DEBUG=y)
 
 Options
+  -c, --configdir   Set configuration directory
   -m, --menuconfig  Open configuration editor
   -h, --help        Help text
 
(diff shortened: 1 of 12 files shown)


dliviu (approved):

example/bootstrap_androidbp.bash
```diff
@@ -50,13 +51,18 @@ EOF
 }
 
 MENU=0
-PARAMS=$(getopt -o hm -l help,menuconfig --name $(basename "$0") -- "$@")
+PARAMS=$(getopt -o c:hm -l configdir,help,menuconfig --name $(basename "$0") -- "$@")
```
kilroyd, line 54: Since we have "c:", this takes an option. configdir should also take an option, so should be "configdir:"

This applies to the other bootstrap files in this review too

kilroyd (commented):
I spotted an issue in the long option handling, but otherwise this should be ok

kilroyd (approved):

PR #464, opened 2020-11-24
'host_bin's shared libraries as implicit dependencies for generator modules

birunts:
After introduced shared library .toc linking functionality, binary
for source generators ('host_bin' property) is not relinked if just
the shared libraries are updated leaving APIs unchanged (.toc unchanged).
Thus shared libraries of used binaries must be added as implicit
dependencies of the generator modules.

Add test case which covers this case.

Change-Id: Ic5268a56610dc46d806273bb1855a0f6fc6f8412


kilroyd (approved):
LGTM

kilroyd (approved):

PR #470, opened 2020-12-11
Fix bob_kernel_module add_to_alias

kilroyd:
Kernel modules should be able to extend existing aliases by using the
add_to_alias property. When we moved the common properties out
BuildProps, we missed implementing getAliasList on kernelModule.

While here, add interface checks to ensure that if interfaces change,
the appropriate modules will get updated.

Change-Id: Ia41cc3b32fdb9423c86199e3f02ae35e6c81029c

### DIFF
--- a/core/generated.go
+++ b/core/generated.go
@@ -154,13 +154,18 @@ type generateCommon struct {
 	}
 }
 
-// generateCommon support {{match_srcs}} on some properties
+// generateCommon supports:
+// * feature-specific properties
+// * module enabling/disabling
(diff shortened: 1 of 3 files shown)


birunts (approved):
LGTM, Thanks!

kilroyd (commented):
Travis isn't running checks in this repo at the moment. We need to get that fixed.
For this commit, I'm happy that we shouldn't be breaking anything on the different versions of go. Sign-off is present. Copyright dates correct. gofmt has been run.

PR #473, opened 2021-01-05
Allow pkg-config binary to be specified

kilroyd:
This may be necessary if a certain pkg-config needs to be used for the
platform. For example there are Debian arch specific pkg-config
packages where the binary is aarch64-gnu-linux-pkg-config.

If necessary an absolute path can be used for the binary.

Change-Id: I73098158182323e7e2cee80e6eaa03420219f52c
Reported-by: Fei Shao <fshao@chromium.org>

### DIFF
--- a/docs/user_guide/libraries_3.md
+++ b/docs/user_guide/libraries_3.md
@@ -87,6 +87,9 @@ bob_binary {
 }
 ```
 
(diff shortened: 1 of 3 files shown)


birunts (approved):
LGTM, Thanks!

PR #475, opened 2021-01-05
Fix kconfig header lookup

francois-berder:
kconfig header could not be found because we were giving an incorrect path to copy_with_deps.


birunts (approved):
Thanks

PR #480, opened 2021-01-12
config_system: Fix integer default handling

kilroyd:
When a configuration value is an integer, has a prompt, but hasn't
been set "by the user" we were setting the value as a string when
being read from a file. This breaks handling of defaults (and other
value comparisons).

Make sure that the value is converted to an integer.

Change set_config() to warn about all values that are detected with
the wrong type.

Change-Id: I011155324fca03ddb51da57a9ca197cea7ade524

### DIFF
--- a/config_system/config_system/general.py
+++ b/config_system/config_system/general.py
@@ -1,4 +1,4 @@
-# Copyright 2018-2020 Arm Limited.
+# Copyright 2018-2021 Arm Limited.
 # SPDX-License-Identifier: Apache-2.0
 #
 # Licensed under the Apache License, Version 2.0 (the "License");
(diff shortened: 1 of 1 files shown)


config_system/config_system/general.py
```diff
@@ -491,24 +491,23 @@ def set_config(key, value, is_user_set=True):
...
-        elif c['datatype'] == 'int':
-            # Must convert to an integer
-            try:
-                value = int(value)
-            except ValueError:
-                logger.warning(
-                    "Ignoring integer configuration option %s with non-integer value '%s'" % (
-                        key, value))
-                return
+    # Validate input
```
chrisdiamand, line 494: Cool - so this now validates all options, not just user_set ones. Makes sense, and I'm surprised we weren't doing this already.

chrisdiamand (commented):
LGTM, only thing is the github checks complaining about the copyright year :/

chrisdiamand (approved):
Thanks!

PR #486, opened 2021-03-15
Soong compatibility layer

chrisdiamand:
A recent change in Soong (`0b0e1b980 AndroidMkEntries() returns multiple
AndroidMkEntries structs`) broke the build of Bob's Soong plugins.

The fix itself is simple - it's just another argument to a callback
function. However, we need compatibility with multiple Android versions
so cannot just make the obvious change.

Instead, add a new `soong_compat` module, which compiles different
helper functions, returning the correct types, depending on the Soong
commit. Infrastructure is added to allow adding more layers in future,
although it is hoped that this will not be necessary.

Detection of the Android version first will do a simple comparison on
ANDROID_PLATFORM_VERSION, which in most cases is enough to uniquely
identify the required source file. If that fails, it will try and do a
comparison using Soong's git SHA, before reverting to the most recent
compatibility file for that particular Android version and printing a
warning.

Change-Id: Ibfa76641deb20b7e9d5ff68a380cf2d7f3aecfba

### DIFF
--- a/core/androidbp_backend.go
+++ b/core/androidbp_backend.go
@@ -1,5 +1,5 @@
 /*
- * Copyright 2020 Arm Limited.
+ * Copyright 2020-2021 Arm Limited.
  * SPDX-License-Identifier: Apache-2.0
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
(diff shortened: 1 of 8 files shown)


kilroyd (approved):
LGTM

PR #488, opened 2021-03-25
Remove ${(mod)_dir} support

kilroyd:
We would like to use Soong's genrule to support the functionality
provided by bob_generate_source and bob_transform_source. In order to
do this, we need to remove some of the functionality currently present
in these modules.

${(mod)_dir} allows access to the intermediate directories of
dependencies. This would allow an undeclared output to be referenced.
It's preferable to avoid this. This affects:

* modules that only want to use particular outputs of a given
  dependency. Workaround: script needs to handle input filtering

* modules that want to pass the intermediate directory to their script
  as an extra include path. This is the use case of wrapping a call to
  a compiler. Workaround: script needs to calculate include paths from
  all outputs

Change-Id: Ie7d0b6d4b6c192d2b4cdb031d7585ffeaf5819d8


birunts (approved):
LGTM, Thanks!

PR #492, opened 2021-04-16
Add support for Fuchsia OS

jbauman42:
Signed-off-by: John Bauman <jbauman@google.com>

### DIFF
--- a/mconfig/basics.Mconfig
+++ b/mconfig/basics.Mconfig
@@ -31,6 +31,9 @@ config OSX
 config WINDOWS
 	bool "Windows"
 
+config FUCHSIA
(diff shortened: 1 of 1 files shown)


dliviu:
Hi,

Thanks for the patch and sorry for the delay in replying. I'm a bit puzzled how this patch adds support for Fuchsia OS given that it is just one entry in the config. Are you planning on submitting future patches that will add more support for Fuchsia OS in Bob?

dliviu (approved):
Apologies for the delay in reviewing this, but as you have the honour of having the first external contribution to this project I had to do the due diligence and cross ts and dot is.

PR #497, opened 2021-08-04
Correct the check for Kconfig file

birunts:
While building kernel modules bob checks if for currently
selected architecture there is a corresponding Kconfig file
in its 'arch' directory which is not the case when the
kernel was built with different output directory
(make O=kernel_out_dir).

For that case there will be a 'source' link created inside
kernel output directory which points to the kernel sources.

With an access to kernel sources Bob should check for
Kconfig file for a corresponding arch.

Change-Id: I7ca9676b76fedfd757773f4065744dbf96d8c4ae


chrisdiamand (approved):
Seems OK to me, thanks!

PR #513, opened 2022-01-21
Add explicit quotes for config variables

birunts:
When a config variable is passed with possible multiple params
e.g. CFLAGS=-g -O0 -DTEST
there are passed to next update_config.py script as separate ones
  1. CFLAGS=-g
  2. -O0
  3. -DTEST

Add explicit quotes for such variables to parse them as single
input parameter.

Change-Id: Id10f7edfb25f72a26c77ba032535cd0ba5089d34

### DIFF
--- a/config.bash
+++ b/config.bash
@@ -1,6 +1,6 @@
 #!/bin/bash
 
-# Copyright 2018-2021 Arm Limited.
+# Copyright 2018-2022 Arm Limited.
 # SPDX-License-Identifier: Apache-2.0
 #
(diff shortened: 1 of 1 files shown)


chrisdiamand (approved):
Seems OK to me!

For the record - this looks strange because everything's already quoted, but we need to use escaped single quotes here to work around the effects of the `eval` on line 48, which we can't remove because it breaks `BOB_CONFIG_PLUGIN_OPTS`.

PR #515, opened 2022-03-09
Add support for A13 Soong

lukokr-aarch64:
Since 6301c3c, Soong now generates install
targets for Make. This clashed with genrulebob
targets and caused cyclic dependencies for install.

The module output will now be correctly set to the
intermediate file and the install rules will be
generated automatically by Soong.

Change-Id: Ibd789c0ac716b33e62b4fc778d03367f1cc558de


birunts (approved):
LGTM! Thanks!

PR #516, opened 2022-04-04
Support nested groups within a list

birunts:
To allow generate native Android's `genrule` and `gensrcs`
`dists` property we need to support a list of
groups which will be rendered as:

 gropupListName: [
     {
          key1: value1,
     },
     {
          key2: value2,
     },
 ]

Group list cannot have inner list. Only group is able to
have group list.

Additionally add support for integer values to handle
gensrcs`s `shard_size` property.

Change-Id: I567403ee1914079a08b3d30aedc37849c4a30722


lukokr-aarch64 (approved):
LGTM, thanks!

PR #517, opened 2022-04-05
Remove implicit_outs from generate/transform source

birunts:
Prepare to migrate to Soong's native `genrule` module by removing
support for implicit outputs on generate/transform source modules.

Note that the 'implicit outs' feature is not removed entirely as
it is still valid and used within other modules.

Change-Id: I64c68939c843f005474aa28e4132d03e0578aca9


lukokr-aarch64 (approved):
LGTM, thanks!

PR #528, opened 2022-09-22
add bob_genrule

BrandonThomasJonesARM:
To be more compatible with android,
we are transitioning to bob_genrule
and away from bob_generate_source.
It is a 1:1 mapping of some Android
genrule properties requiring only a
linux backend generator.

Signed-off-by: Brandon Jones <brandon.jones@arm.com>


lukokr-aarch64 (approved):
LGTM, thanks !

PR #529, opened 2022-09-26
Refactor `phonyInterface`

birunts:
As `bob_resource` module implements `phonyInterface` it has to define `outputs()` and `implicitOutputs` even though it doesn't have any. Both methods need to be moved to `dependentInterface`.

Change-Id: Iad188d6cc3f0a0fa472f4a8868e1620cafc2d680


BrandonThomasJonesARM (approved):
+2 provided tests pass :)

PR #552, opened 2023-02-13
fix(ci): reduce number of macos jobs

lukokr-aarch64:
The MacOs jobs take a long time and have less
agents available. A single job should be
enough coverage for our needs.

Change-Id: If24a65c60f321871f76a789423bf7250391dfaa9


birunts (approved):
Good catch, thank you.

PR #559, opened 2023-02-23
feat(gazelle): add config structure

lukokr-aarch64:
`Configure` runs in every directory of the project, the plugin configuration is propagated down the directory tree by inheriting the values from its parent.

The `bob_root` directive is used to mark the root module of the Bob workspace such that the plugin will parse the files in that tree recursively.

This allows the plugin to support multiple Bob workspaces in a monorepo setup.

Change-Id: I310083b582bfbca0d9d1754486769502d4cf6d35


birunts (approved):
:+1:

PR #607, opened 2023-05-17
fix(toolchain): fixes search and replace bug

lukokr-aarch64:
Change-Id: If030216f39de588de620b21279ad1d436941759f


birunts (approved):
Good spot! Thanks!

PR #619, opened 2023-06-02
Flag refactor part1

lukokr-aarch64:


birunts (approved):
Nice piece of work! Thanks!

PR #660, opened 2023-08-31
chore: add convience snapshot script

lukokr-aarch64:
The `:verify_snapshots` target does not work as expected currently. For now, remove it and provide a shell
script to easily update any failing snapshots.

This is useful when mass adding large amounts of
test cases and saves time by not running the update on passing test cases.

Change-Id: Ib91d1f0be2f7f5afabb0776102cfec602ef89855


birunts (approved):
Thx!

PR #669, opened 2023-10-06
docs: reformat `bob_alias` docs

lukokr-aarch64:
Added docs for common properties for both legacy and new modules. This will allow us to be clearer about what is supported by both.

Change the style of attribute documents to clearly mark types of attributes and their behaviour (required, default, etc).

Generally if the description of the field is a single paragraph then document this directly, otherwise link to a detail page.

Change-Id: I1165908a8b4c9f2027ef75feb3a537dc6cffd125


birunts (approved):
Awesome! Thank you for the work done!

PR #696, opened 2024-08-19
On-demand feature struct allocation

chrisdiamand:
Bob's conditionals in modules ("features") can use significant amounts of RAM during build.bp parsing - each config option spawns a new struct containing a copy of each build property, resulting in a multiplicative impact on memory consumption.

Fortunately, close inspection of Blueprint's "unpack" code reveals that we can instead pass in a nil pointer to the struct, instead of initialising the properties for each conditional feature, which will cause property structs to be allocated only when actually used in the build.bp file.

Use this in Bob by making the property struct for each feature a nil pointer to the property struct type, instead of a fully-instantiated interface{} wrapped in a BlueprintEmbed.

Change-Id: I42e432fcaa8f1f5cd48ba01c53cb95ce5109aa07


core/feature.go
```diff
@@ -71,20 +67,14 @@ func (f *Features) Init(properties *config.Properties, list ...interface{}) {
 	for i, featureName := range properties.FeatureList {
 		fields[i] = reflect.StructField{
 			Name: featurePropertyName(featureName),
-			Type: reflect.TypeOf(singleFeature{}),
+			Type: reflect.PointerTo(propsType),
```
kilroyd, line 70: I guess we need to use PtrTo if we're still supporting go versions before 1.18

https://go.dev/pkg/reflect/?m=old#PointerTo

birunts (approved):
Thanks!

PR #704, opened 2024-09-30
fix: bring back `set_config_if_prompt`

birunts:
There is a need to fall back `set_config_if_prompt` function in `system_config` module as deprecated. This is to allow easier transition to the newer API. 

Bob tests does not check if allowed plugins can properly work with config system. Change also adds simple test plugin which sets config option and verifies that.


jonkirkham-arm (approved):
LGTM

PR #716, opened 2025-11-29
replace EXTRA_CFLAGS with CFLAGS_MODULE

pleasurefish:
fix the build issue with linux kernel 6.18.
The EXTRA_CFLAGS setting is outdated, causing some directives previously passed through EXTRA_CFLAGS to be omitted, which resulted in certain definitions becoming undefined.

### DIFF
--- a/scripts/kmod_build.py
+++ b/scripts/kmod_build.py
@@ -44,11 +44,11 @@ def parse_kbuild_options(kbuild_options):
 def check_kbuild_option_conflicts(kdir, option, value):
     """
     Checks whether kbuild_option is already defined in the kernel config
-    causing conflicts by overriding them in EXTRA_CFLAGS
+    causing conflicts by overriding them in CFLAGS_MODULE
     :return: Zero(0) if there is no conflict, one(1) otherwise
     """
     message = (
-        "Overridden '{0}' option in EXTRA_CFLAGS. "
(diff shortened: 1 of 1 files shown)


pleasurefish:
gently ping.

BrandonThomasJonesARM:
hi @pleasurefish sorry for the long time to respond.

This project is beginning to be deprecated and we are mainly working exclusively on ensuring our teams internal transition away from bob is smooth. We are removing support from building kernels from bob altogether so we are unlikely to support this.

pleasurefish:
@BrandonThomasJonesARM , so what's your next? how do you build the kernel without bob?

jonkirkham-arm:
For Android use [Kleaf](https://android.googlesource.com/kernel/build/+/refs/heads/main/kleaf/README.md?pli=1#), and for Linux use [Kbuild](https://docs.kernel.org/kbuild/index.html).

pleasurefish:
[jonkirkham-arm](https://github.com/jonkirkham-arm), thanks for your information.
Bob‑build is able to build both the userspace library and the kernel module for the Mali DDK in a single workflow.
Does this mean there will be another tool that calls into Kleaf or Kbuild to provide the same capability?

jonkirkham-arm:
There are no plans for another tool which can user + kernel builds together. Tt means you must run Kbuild or Kleaf separately to build kernel Modules.

